PR #32, opened 2020-06-16
Update boards.bml

RokkumanX:
This does not only update the boards.bml to match the latest one from higan (2020-01-21), I also added entries SHVC-1L5B-02 (Super Mario RPG Japan) and SHVC-BA3M-20 (Estpolis Denki II Japan)

I also bumped the revision dates.


RokkumanX:
It doesn't look like the boards.bml is fully updated with this commit/pull request.

Super Mario RPG Japan and Estpolis Denki II Japan does not work at all, I can see that the labels are correct and that they are flagged with verified status but bsnes seems to read from the old boards.bml file so the new changes aren't read.

Issue #19, opened 2020-04-24
Support integrating newer revisions of Sameboy as they release.


Yersho:
I don't know if you are already doing this or not but I was wondering if you plan on integrating future revisions of Sameboy into the emulator for it's Super Game Boy emulation. Sameboy is only on revision 0.12.3 so it would make sense to update BSNES's Sameboy core as new releases come out.

Screwtapello:
It looks like currently [`bsnes/bsnes/gb/`](https://github.com/bsnes-emu/bsnes/tree/30e274953a8d7dc10e71be0987c15018e397df2f/bsnes/gb) is a manually-stripped-down copy of [SameBoy's source structure](https://github.com/LIJI32/SameBoy/tree/ddad913e06fb51e80957177d404fef3d1d3b5d6c).

In *theory* we should be able to just make that directory a git subtree of SameBoy and pull updates whenever we like... but it *would* be messier and more confusing to have the entire SameBoy source-tree and build-system included when we use hardly any of it. I'm also not entirely sure how to *test* bsnes after updating SameBoy, other than just "does Donkey Kong '94 run".

Screwtapello:
To actually answer your question, though: I would like to keep bsnes' bundled version of SameBoy up-to-date, it's just a matter of figuring out what needs to be done, and finding the time to do it.

LIJI32:
0.12.3 is the latest stable version of SameBoy, bsnes shouldn't use a non-stable version (As the save state format varies between revisions, and stable releases expect save state from other stable releases). Also, byuu and I took care of making sure all of bsnes' changes to SameBoy are either reintegrated or reimplemented in a way that does not require changes to the core, so updating SameBoy is just a matter of copying a new `Core` directory from SameBoy over the old one.

RokkumanX:
bsnes is currently updated with SameBoy 0.13.1 and nothing overly important have happened after, is that enough?

LIJI32:
0.13.2 fixed a few bugs related to save states. Specifically, it fixed a case where save states from Windows did not work on other OS and vice versa.

RokkumanX:
Glad to see some response, I thought everyone died.

So updating the core isn't that hard right? Maybe I should attempt it and see what happens.

RokkumanX:
OK I made some commits but I couldn't find anywhere to change 0.13.1 to 0.13.2

You should be able to change files and only make one commit, but that's beyond my knowledge.

PR #42, opened 2020-06-24
Update SameBoy 0.13.2

RokkumanX:
Should be a proper Pull Request this time

### DIFF
--- a/bsnes/gb/Core/debugger.c
+++ b/bsnes/gb/Core/debugger.c
@@ -689,6 +689,7 @@ exit:
 
 struct debugger_command_s;
 typedef bool debugger_command_imp_t(GB_gameboy_t *gb, char *arguments, char *modifiers, const struct debugger_command_s *command);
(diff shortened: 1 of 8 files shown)


Kawa-oneechan:
> ./gb/Core/gb.h:503:13: error: 'GB_gameboy_s::<unnamed union>::<unnamed struct>::<unnamed enum> GB_gameboy_s::<unnamed union>::<unnamed struct>::GB_FRAMESKIP_SECOND_FRAME_RENDERED' invalid; an anonymous struct may only have public non-static data members [-fpermissive]

Yeah I don't think this is fit to merge.

RokkumanX:
> so updating SameBoy is just a matter of copying a new Core directory from SameBoy over the old one.

Seems to me like it wasn't as easy as stated in #19 
I only updated the files in the Core folder that was affected by 0.13.2 and nothing else.

I did my best trying to be helpful, now I don't really care anymore.

Good luck!

PR #47, opened 2020-07-26
GNU Make 4.3 compatibility

Alcaro:
```
* WARNING: Backward-incompatibility!
  Previously appending using '+=' to an empty variable would result in a value
  starting with a space.  Now the initial space is only added if the variable
  already contains some value.  Similarly, appending an empty string does not
  add a trailing space.
```
https://lwn.net/Articles/810071/

Should be applied to all copies of nall across both @bsnes-emu and @higan-emu, as well as Near's upstream (do you have his contact info?).

### DIFF
--- a/nall/GNUmakefile
+++ b/nall/GNUmakefile
@@ -2,13 +2,14 @@
 MAKEFLAGS := Rr
 .SUFFIXES:
 
+# An empty variable to defeat make functions that trim whitespace.
+[empty] =
 [0-9] = 0 1 2 3 4 5 6 7 8 9
 [A-Z] = A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
 [a-z] = a b c d e f g h i j k l m n o p q r s t u v w x y z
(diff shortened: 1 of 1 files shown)


Screwtapello:
So far as I can tell, the only place `$([space])` is used is in hiro's and ruby's `make verbose` rules:

https://github.com/bsnes-emu/bsnes/blob/24503fc9221a4b3b796eb94d645cbbc63f207859/hiro/GNUmakefile#L79-L81

The end result is that on Make 4.3, hiro's backend and the list of ruby drivers are indented by only a single space instead of two spaces as in previous versions (thanks to Hendricks266 on Discord for verifying the output on an old version of Make for me). Not the most dramatic of problems, but worth fixing.

Screwtapello:
Sorry it took so long to review your PR, but thank you very much for making it! I've submitted your changes to higan as https://github.com/higan-emu/higan/pull/112 as well.

PR #49, opened 2020-07-29
Make the README use local assets, not remotely-hosted ones.

Screwtapello:
When Near took down byuu.org, he also removed the images that the bsnes README was ointing to. He gave me backups of data from the site, including the images used on the bsnes homepage, but apparently the README used different images hosted in a different location.

I've replaced the missing images with ones from the bsnes homepage that look reasonable.

[Rendered](https://github.com/Screwtapello/bsnes/blob/61223adb9a05c1d668e072a6c149548a8d4124ba/README.md)


Kawa-oneechan:
Looks good to me

PR #50, opened 2020-07-31
Update `Super Famicom.bml`

n3vu0r:
First commit updates `Super Famicom.bml` to the one of ares v115.

Second commit adds the official Star Fox 2 from Super NES Classic Edition. Since all Super NES Classic Edition games have USA localization, I chose `region: USA` and `revision: 1.0` like it's done with prototype boards. Since there is no physical board, I chose the generic `board: GSU-RAM` which is suggested by bsnes's heuristics. In the end this board definition is identical to the board `SHVC-1CA6B-01` used by Stunt Race FX which is considered a good donor cartridge. Does this make sense?

### DIFF
--- a/bsnes/Database/Super Famicom.bml
+++ b/bsnes/Database/Super Famicom.bml
@@ -1,5 +1,5 @@
 database
-  revision: 2020-01-01
+  revision: 2020-01-21
 
 //Prototypes (JPN)
 
@@ -125,7 +125,7 @@ game
 //Super Famicom (JPN)
 
 database
-  revision: 2020-01-01
+  revision: 2020-01-21
 
 game
   sha256:   5c4e283efc338958b8dd45ebd6daf133a9eb280420a98e2e1df358ae0242c366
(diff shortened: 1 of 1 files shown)


Screwtapello:
Thank you very much for this PR, and thank you for taking the time to figure out how bsnes' database needed to be updated.

I have rebased your updates-from-ares change and committed it as 9ef085fb.

I haven't included the Star Fox 2 change because I haven't figured out what the rules for adding to or updating the game database should be, going forward. Right now, a game's entry in the database is a strong signal that a particular game was intended to run on a particular board, but since Star Fox 2 was never sold on a board it's hard to say exactly what board it was intended to use. *Probably* it was `SHVC-1CA6B-01` or something very similar, but we'll never know for certain. Since Star Fox 2 seems to work perfectly well with the heuristically-generated manifest, I don't think there's a pressing need for it to be in the database.

Feel free to re-open this PR, or open an issue, if there's some detail I've missed!

n3vu0r:
Makes sense, thanks for explaining.

PR #55, opened 2020-08-15
Consider the overscan setting when getting the aspect ratio

hizzlekizzle:
I'm looking to move the libretro fork's parent over to this repo, since the original parent is now gone and I assume this is where the action will be taking place from here on out. This is the one commit we had locally that wasn't upstream, so I figured I'd offer it up: https://github.com/libretro/bsnes/commit/f04fe9bf5bfe8fc6a3ce3ee2c90747daf579258c

### DIFF
--- a/bsnes/target-libretro/libretro.cpp
+++ b/bsnes/target-libretro/libretro.cpp
@@ -46,20 +46,27 @@ static int aspect_ratio_mode = 0;
 
 static double get_aspect_ratio()
 {
+	double ratio;
+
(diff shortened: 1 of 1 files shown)


Kawa-oneechan:
Looks good to me.

Screwtapello:
Thanks for the patch, @hizzlekizzle!

hizzlekizzle:
Np, thanks for the merge! Moving forward, I'll archive our old repo and just make a new fork for us based on this upstream.

PR #67, opened 2020-09-30
Merge upstream libco changes

rtretiakov:
This merges upstream libco changes and brings in a script based on the one in higan-emu/higan to update with any future changes.


Screwtapello:
Thank you very much for looking into this!

Unfortunately this PR was made from the master branch *before* your PPU mosaic fixes PR landed, and trying to rebase a merge like this is... difficult.

Instead, I have cherry-picked your "add update-subtrees.sh" commit, and then ran that script to create a new subtree merge, which I then pushed.

PR #68, opened 2020-10-02
libretro: Add support for BS-X Satellaview

rtretiakov:
This adds support for the BS-X Satellaview to the libretro port. It works similarly to Super Game Boy.

### DIFF
--- a/bsnes/target-libretro/program.cpp
+++ b/bsnes/target-libretro/program.cpp
@@ -15,6 +15,7 @@ using namespace nall;
 #include <heuristics/heuristics.cpp>
 #include <heuristics/super-famicom.cpp>
 #include <heuristics/game-boy.cpp>
+#include <heuristics/bs-memory.cpp>
 
 #include "resources.hpp"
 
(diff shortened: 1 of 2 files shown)


bsnes/target-libretro/program.cpp
```diff
@@ -506,6 +548,24 @@ auto Program::loadGameBoy(string location) -> bool {
...
+	string manifest;
+	vector<uint8_t> rom;
+	rom = loadFile(location);
+
+	if (rom.size() < 0x8000) return false;
+
+	auto heuristics = Heuristics::BSMemory(rom, location);
+	auto sha256 = Hash::SHA256(rom).digest();
+
+	bsMemory.manifest = manifest ? manifest : heuristics.manifest();
```
Screwtapello, line 561: Shouldn't this just be `bsMemory.manifest = heuristics.manifest();` ?

rtretiakov: Yes.

PR #69, opened 2020-10-09
Update SameBoy to 0.13.6

rtretiakov:
This updates SameBoy to the 0.13.6 release, with a build fix backported from prerelease.
The .inc files that were deleted in the diff were stale, and had been moved into the graphics/ subdirectory already, but somehow stayed alongside the source files in previous updates.


Screwtapello:
Hmm.. to make updating easier in future, perhaps SameBoy should be subtree'd in and added to [update-subtrees.sh](https://github.com/bsnes-emu/bsnes/blob/0ab0ab69572a88be6dcc633662d5e2cda807ea45/update-subtrees.sh)?

rtretiakov:
Here are my thoughts on that, collected here so as not to lose them:

We need to consider whether updates should be done when SameBoy releases are tagged, or if merging the master branch at arbitrary times is the plan. If we want to go with tagged releases, then using a script needs to take that into account, by either taking a command line argument or reading through the tags and selecting the latest one in the SameBoy repository.

I've manually deleted files and backported a fix. I also updated presentation.cpp to show the new version number. If we merge now using subtrees, then those things will need to be reapplied.

So I think merging only tagged SameBoy releases, and merging this now and planning to use a scripted method for the next tagged release makes the most sense.

Screwtapello:
I've created PR #71 with the subtree approach. I haven't deleted unused files, and the `update-subtrees.sh` script just always grabs the latest version. I'm not *too* worried about that, since I don't really expect the script to be run regularly, more like used as a reference to copy/paste commands from.

I'm not married to the idea, though, which is why I made it a PR rather than just landing the change. I'll canvas the Discord community to see if anybody else has any strong preferences.

Either way, thank you for hounding me to make this update, which hopefully would have happened eventually, but wouldn't have happened nearly as quickly without your encouragement.

Screwtapello:
I went ahead and merged #71 instead of this PR. Let me know if you have any problems.

PR #71, opened 2020-10-12
Sameboy subtree

Screwtapello:
This is an alternative to #69. Compared to that PR (manually merging changes from SameBoy):

  - this includes all of SameBoy, including all the bits we don't use or care about
  - this makes it easier to update (just run `update-subtrees.sh`) but harder to update to a specific release (`update-subtrees.sh` always pulls the latest version, you'd have to run the subtree command manually to grab a specific revision)
  - this makes it easier to contribute changes upstream (we can peel that subtree back out to get commits we can push as PRs)

Like #69, this includes some fixes that have landed since 0.13.6, so that version number is not *strictly* accurate, but close enough.


rtretiakov:
I'm not attached to either pull request being merged, but either way this work needs to be done for the next SameBoy release. Nothing has been done around tags here but I don't think that is super important right now. We can add it later if it becomes an issue.

It does appear the stale .inc files are also not included in bsnes/gb/Core here.

As far as I can tell from a first glance it looks good.

Screwtapello:
After a couple of days, nobody has tried very hard to change my mind, so I went ahead and merged this.

PR #76, opened 2020-10-20
Strip whitespace and empty lines from the cheat editor.

### LINKED ISSUE #63: Error when adding new cheat.
I am using the latest nightly build.

Open any ROM.
Select Tools/Cheat Finder...
Press scan
Double click on any of the finds.
Enter a name and click Add.
You see wrong error message.
Remove empty line and press Add.

Problem is that that line does not filter out empty lines and then ``decodeSNES()`` reports an error.

https://github.com/bsnes-emu/bsnes/blob/a8c1f19f7512e38b73620875145b86e251721a0e/bsnes/target-bsnes/tools/cheat-editor.cpp#L104

Screwtapello:
Previously, bsnes would turn the cheat text into a list of cheats by splitting on '\n', producing an empty "cheat" after the last line of the editor:

    "7e0003=00\n".split("\n") -> ["7e0003=00", ""]

That empty "cheat" is not a valid cheat, so the result would be marked invalid. Teaching bsnes to silently skip empty lines worked, but it would preserve multiple blank lines between two cheats, which was silly.

Now we iterate over all the lines of the cheat editor, putting valid codes in a new vector, and setting a flag for invalid codes.

This fix works for me, but I would particularly appreciate code-review from people who understand C++ memory-management - in Rust you can't move items from one vector to another unless iterating over the old vector consumes it and will release its memory; it's not clear to me whether nall's vector iteration is consuming or read-only, or whether the item is copied or moved to the new vector. It *probably* doesn't matter, since both vectors are local variables and the only storage to escape the function is a freshly-allocated string regardless, but I'd like to know for future reference.

Fixes #63.

### DIFF
--- a/bsnes/target-bsnes/tools/cheat-editor.cpp
+++ b/bsnes/target-bsnes/tools/cheat-editor.cpp
@@ -101,11 +101,17 @@ auto CheatWindow::doChange() -> void {
 }
 
 auto CheatWindow::doAccept() -> void {
-  auto codes = codeValue.text().downcase().transform("+", "\n").split("\n").strip();
+  auto raw_codes = codeValue.text().downcase().transform("+", "\n").split("\n").strip();
+  vector<string> valid_codes;
   string invalid;  //if empty after below for-loop, code is considered valid
-  for(auto& code : codes) {
+  for(auto& code : raw_codes) {
+    if(code.length() == 0) {
+      continue;
+    }
     if(!program.gameBoy.program) {
-      if(!cheatEditor.decodeSNES(code)) {
+      if(cheatEditor.decodeSNES(code)) {
+        valid_codes.append(code);
+      } else {
         invalid =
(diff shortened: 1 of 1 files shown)


Alcaro:
It's hard to say anything 100% certain when there's so much auto, but that looks safe to me, just looks like it's iterating references to the vector contents (mutable references, but who cares - C++ doesn't) and inserting copies.

If you want to move it, use valid_codes.append(std::move(code)), though the performance improvement is most likely insignifiant (and I'm not sure if nall::string supports being moved - if it doesn't, std::move just reverts to copy).

Screwtapello:
Thanks for the review, Alcaro!

I tested this branch with Clang's Address Sanitizer, and it was happy. I also tried it under Valgrind, and Valgrind hit its limit of 1000 reported errors before I even managed to open the Tools menu, never mind actually testing this change.

I'm satisfied.

PR #77, opened 2020-10-20
nall: Use unprefixed endian defines from endian.h.

niacat:
From what I can tell every system with `<endian.h>` has `BYTE_ORDER`, `BIG_ENDIAN`, and `LITTLE_ENDIAN` without a double-underscore __prefix.

* [macOS (best I could find)](https://opensource.apple.com/source/xnu/xnu-4570.71.2/bsd/arm/endian.h.auto.html)
* [FreeBSD](http://bxr.su/FreeBSD/sys/arm/include/endian.h#54)
* [NetBSD](http://bxr.su/NetBSD/sys/sys/endian.h#94)
* [OpenBSD](http://bxr.su/OpenBSD/sys/sys/endian.h#43)
* [POSIX](https://www.austingroupbugs.net/view.php?id=162#c665)
* [glibc libc](https://code.woboq.org/userspace/glibc/string/endian.h.html#44)
* [musl libc](https://github.com/cloudius-systems/musl/blob/master/include/endian.h#L16)

On Linux, define these macros to the double underscore variants if they're missing, just in case they're not defined properly on older versions.

On Windows, I think it's covered by the `defined(_M_IX86) || defined(_M_AMD64)` part.
 
This fixes endian detection and lets bsnes build on NetBSD/aarch64, which does not have the prefixed defines.

### DIFF
--- a/nall/intrinsics.hpp
+++ b/nall/intrinsics.hpp
@@ -121,6 +121,15 @@ namespace nall {
   #include <machine/endian.h>
 #elif defined(PLATFORM_LINUX)
   #include <endian.h>
+  #ifndef BYTE_ORDER
+    #define BYTE_ORDER __BYTE_ORDER
+  #endif
+  #ifndef LITTLE_ENDIAN
+    #define LITTLE_ENDIAN __LITTLE_ENDIAN
+  #endif
+  #ifndef BIG_ENDIAN
+    #define BIG_ENDIAN __BIG_ENDIAN
+  #endif
 #elif defined(PLATFORM_BSD)
   #include <sys/endian.h>
 #endif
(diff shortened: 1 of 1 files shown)


Screwtapello:
Here's how I understand what's going on.

Pre-processor defines are about the only way to configure C/C++ code at build time, so compiler and libc authors want to make lots of pre-processor defines to advertise the features they offer.

However, application authors also want to name things in their applications without worrying about whether some future compiler or libc is going to define a conflicting name.

So, the C and C++ committees decided that any identifier with a leading underscore, or containing consecutive underscores, was reserved for the implementation, while any name without those features was reserved for applications. So you can name something `fred` or `FRED` without worrying that somebodying going to make a CPU called "Fred" and C compilers targeting that CPU will add a pre-processor definition that breaks your code.

However, C only got standardised well after it was already popular, and there were a lot of legacy definitions that didn't use underscores when (according to the new standard) they should have. Like `LITTLE_ENDIAN`.

So, compiler and libc vendors tended to include those defines with standards-compliant names (`_LITTLE_ENDIAN` or `__LITTLE_ENDIAN`) and only adding the legacy names when you explicitly opt into legacy, non-standard content. In glibc and musl, you do that by defining `_GNU_SOURCE` or `_BSD_SOURCE` (depending on which variants you want) before including anything; it looks like macOS provides those things *unless* you ask for `_POSIX_C_SOURCE`, FreeBSD requires `__BSD_VISIBLE`, NetBSD requires `_XOPEN_SOURCE` or `_NETBSD_SOURCE`.

I see that the no-underscore names have been added to POSIX, but it's not clear to me which version of POSIX contains those changes, and my `endian(3)` glibc manpage doesn't mention them.

So, there's a number of things we could do:

  - we could get a consistent definition by defining `LITTLE_ENDIAN` (etc.) to match `_LITTLE_ENDIAN` or `__LITTLE_ENDIAN`, but the end goal is to consistently define `ENDIAN_LSB` or `ENDIAN_MSB` — adding redundant consistent definitions seems like a waste.
  - we could get consistent definitions by adding the "give me legacy content" definitions, but *those* definitions also vary from platform to platform (and as mentioned, I'd rather not add any of those)
  - we could add even more complexity to the existing garguantuan test:

        #if    (defined(__BYTE_ORDER) && defined(__LITTLE_ENDIAN) && __BYTE_ORDER == __LITTLE_ENDIAN) \
            || (defined( _BYTE_ORDER) && defined( _LITTLE_ENDIAN) &&  _BYTE_ORDER ==  _LITTLE_ENDIAN) \
            || (defined(  BYTE_ORDER) && defined(  LITTLE_ENDIAN) &&   BYTE_ORDER ==   LITTLE_ENDIAN) \
            || defined(__LITTLE_ENDIAN__) \
            || defined(__i386__) || defined(__amd64__) \
            || defined(_M_IX86) || defined(_M_AMD64)

That last one is kind of a mess, but maybe it's legitimately the least terrible?

Screwtapello:
Oh, also — thank you for working on portability issues! Portable code can be a pain to write, but it lasts longer than the other kind, so it's worth pursuing. We currently use CirrusCI and GitHub Actions for nightly builds, and neither of those support NetBSD, but if there's some free service that can warn us if we introduce a change that accidentally breaks building on NetBSD, perhaps we could hook that up.

Screwtapello:
I've made my own PR based on the approach I mentioned at the end of https://github.com/bsnes-emu/bsnes/pull/77#issuecomment-713358515 — if you can confirm that it fixes NetBSD/aarch64, and it passes all the other CI tests, I'll merge it.

Thanks for helping to make bsnes more portable!

Issue #61, opened 2020-09-19
Merge emulation fixes from higan


Screwtapello:
A few minor emulation fixes have landed in higan since bsnes was forked. We should cherry-pick those fixes across to maintain parity. You can generate a list of candidates in the higan repository with `git log v110.. higan/sfc` (assuming byuu already synced any changes from the fork-point to v110), but as of this writing they include:

- [X] https://github.com/higan-emu/higan/commit/caed1556ebb460cee27589bac7bdf9b0567a88e1
- [x] https://github.com/higan-emu/higan/commit/56bdd466461201f80001cc9db4bded1360e81336
- [ ] https://github.com/higan-emu/higan/commit/e422ddc7293b10aae8a1935d4c7f96aeb8445430

(I'm not sure about that middle one, but it's worth investigating)

Screwtapello:
These fixes may allow us to remove game-specific workarounds in bsnes, too.

Kawa-oneechan:
This gets my vote.

Max833:
Great idea. And we shouldn't forget the mosaic fixes that went into ares but not bsnes. This is the mosaic test (ares is now hardware accurate):
https://github.com/PeterLemon/SNES/tree/master/PPU/Mosaic/Mode5

![image](https://user-images.githubusercontent.com/53529402/93684380-9be63e80-faaf-11ea-8c6c-f18c059b0531.png)

Kawa-oneechan:
Ooh, good call.

Screwtapello:
It's been pointed out to me that https://github.com/higan-emu/higan/commit/caed1556ebb460cee27589bac7bdf9b0567a88e1 is already changed in bsnes, so that's one down, I guess.

rtretiakov:
I've added 56bdd46 to my fork and will be making a pull request.

Screwtapello:
Just going by commit messages, the SNES mosaic fixes were in https://github.com/higan-emu/higan/commit/73b52e4d34b5edde080a32a16ca35cbdced437be and https://github.com/higan-emu/higan/commit/8b493a71960cf43e8b2dfbcddb4bc8a24bdd08a5.

rtretiakov:
Thank you @Screwtapello , I will see about merging those fixes as well as the input one after some more analysis.

Screwtapello:
Thank *you* for looking into it!

rtretiakov:
[...]

PR #78, opened 2020-10-22
Fix auto-joypad polling

asuramaru:
This merges SNES auto-joypad polling fixes. The new code runs the auto-joypad polling at 128-cycle granularity instead of 256-cycle granularity, so that the first event can happen ~128 cycles after the start of Vblank. I've left comments for the unknown behaviors. I also removed the hacks since they are no longer needed.

This fixes issue 3 of 3 for https://github.com/bsnes-emu/bsnes/issues/61, which means we can close issue 61 now.

I have tested and confirmed these fixes work with every game affected.

### DIFF
--- a/bsnes/sfc/cpu/timing.cpp
+++ b/bsnes/sfc/cpu/timing.cpp
@@ -200,7 +200,7 @@ auto CPU::dmaEdge() -> void {
   }
 }
 
-//called every 128 clocks; see CPU::step()
+//called every 128 clocks from inside the CPU::stepOnce() function
 auto CPU::joypadEdge() -> void {
   //it is not yet confirmed if polling can be stopped early and/or (re)started later
   if(!io.autoJoypadPoll) return;
--- a/bsnes/sfc/cpu/timing.cpp
+++ b/bsnes/sfc/cpu/timing.cpp
@@ -5,7 +5,7 @@ auto CPU::dmaCounter() const -> uint {
 
 //joypad auto-poll clock divider
 auto CPU::joypadCounter() const -> uint {
-  return counter.cpu & 255;
+  return counter.cpu & 127;
 }
 
 auto CPU::stepOnce() -> void {
@@ -105,7 +105,7 @@ auto CPU::scanline() -> void {
     status.hdmaSetupPosition = (version == 1 ? 12 + 8 - dmaCounter() : 12 + dmaCounter());
     status.hdmaSetupTriggered = false;
 
-    status.autoJoypadCounter = 0;
+    status.autoJoypadCounter = 33;  //33 = inactive
   }
 
   //DRAM refresh occurs once every scanline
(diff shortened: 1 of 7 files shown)


asuramaru:
I worked on this back when the bug was first uncovered, but since jbo85 beat me to it, I sat on it, but it didn't get ported back to bsnes, so I finished up my approach for this pull request.

My code is not identical to jbo85's, but it's very very close. Both of us are guessing on a few small unknowns regarding early termination / restart of polling within the same frame. I'm obviously biased to my approach, considering it's less complex and needs no new state variables or I/O handler changes, but I think it'll be a very good thing to have bsnes and higan each have a separate implementation. If in the future anyone comes across an input bug, we can compare to the other emulator and if one works, then we know whose approach was right and we've gleaned new knowledge in the process.

I suspect we'll never find out though. The differences are so subtle that even for a system as quirky as the SNES, I don't foresee any games relying on it.

Screwtapello:
I've tested this PR with all the games listed in https://github.com/higan-emu/higan/pull/89, which includes all the games whose per-game hacks were removed. *Taikyoku Igo - Goliath* was also a test-case for https://github.com/higan-emu/higan/commit/e422ddc7293b10aae8a1935d4c7f96aeb8445430.

Everything seems to be working, given a few minutes' poking about.

bsnes/sfc/cpu/timing.cpp
```diff
@@ -200,75 +200,47 @@ auto CPU::dmaEdge() -> void {
   }
 }
 
-//called every 256 clocks; see CPU::step()
+//called every 128 clocks; see CPU::step()
```
Screwtapello, line 203: It seems this is actually called from `CPU::stepOnce()` these days, not `CPU::step()`.

Screwtapello: Also, your PR comment says:

> The new code runs the auto-joypad polling at 128-cycle granularity instead of 256-cycle granularity, so that the first event can happen ~128 cycles after the start of Vblank.

Why is that desirable, if Jonas Quinn's changes to higan didn't need it? (I'm genuinely confused here)

Screwtapello (changes requested):
I confess I find your implementation a bit easier to understand than the one in higan, because it doesn't have that extra state variable. Well done!

Do you think it would be helpful to copy the comments from Jonas Quinn's higan changes into bsnes?

Also, please update your commit message to include the useful and interesting information you put into the PR description. It would be a shame for that extra context to be lost!

To be clear, the changes I want to see before merging are:

  - context in commit message
  - change the reference to `CPU::step()` in a comment

I'd like to learn more about the 256 cycle → 128 cycle change, but that's not a code change, just a personal understanding thing.

Apart from that, it looks good!

Thank you very much for helping to keep bsnes a world-class emulator. :)

Screwtapello:
Also, it seemed a shame to lose the list of "games that are sensitive to auto-joypad polling behaviour", so I stuck it on byuu.net: https://byuu.net/emulation-bugs/snes/

bsnes/sfc/cpu/timing.cpp
asuramaru: We might as well make the comment CPU::stepOnce, good catch!

anomie's notes were slightly off. So let me back up: 32.5-95.5 is referring to the PPU dot clock. If you multiply by 4 you get the cycle count, 130-382. It's pretty clear that in 256-cycle units, that would need to be 130-384 or you'd have no event when it lands on cycle 384. Jonas Quinn confirmed that with his testing in any case.

I chose to split the cycle range in half and run things every 128 clocks. So that reduces us to 32.5-64.0, or 130 to 256 + 258-384. The reason I chose 128 over 256 was so I could emulate the time it takes to perform the joypad latching. I don't believe you can set the controller latch to 1 and then 0 immediately. It needs time to let the controllers see this happening and then release the latch, and time to let that take effect. Polling the controllers is only one operation so it makes sense that happens once every 256 cycles, which is why I have the (counter&1) test for that.

Think about it this way: we know the polling sequence takes 4224 cycles, and 4224/256 is 16.5. Isn't that a little strange to have half a cycle? But if we consider the latching is actually twice as fast, then it's 2x128 + 16x256. If we run on a 128-clock counter, it becomes 2x128 + 16x2x128, or 34 total states. But now since we don't actually poll on the second half of each 16x2x256 states, we don't need to bother running the event there, so we can stop at 33 instead of 34 and it's the same thing. And by doing that, it turns our polling duration into 4224 cycles, just like we've observed.

Jonas Quinn's patch does the latch to 1 then latch to 0 at the same instant, which is why he can use a 256-cycle counter.

asuramaru:
> Do you think it would be helpful to copy the comments from Jonas Quinn's higan changes into bsnes?

Good catch, some of it could be useful here. We don't know what happens when you read 4218-421f or 4016-4017, or write to 4200 during the polling sequence. I'm not sure where those comments belong, but I would guess inside the io.cpp file next to those read/write handler sections. Unfortunately I don't know how to add another file to my change request to put the comments there. But I could make a second pull request after this one to add those comments if you like. Or if you can do that, I can point you to line#s and comments to add.

Also not to nitpick but for the higan commit, "   // Todo: What happens to auto-joypad read when you read/write 4816/4817 while it is in progress?", that's a typo, I'm pretty sure he meant to say 4016/4017 since 4816/4817 aren't registers.

"// Todo: 4224 cycles would be 33*128 and not 16*256, hardware testing shows that higan ends auto-joypad read too early", ah I see he was on the same page as I was here. Great minds? Hahah!

> Also, please update your commit message to include the useful and interesting information you put into the PR description. It would be a shame for that extra context to be lost!

I made your change to the stepOnce comment, good catch there!

I'm sorry to say but I don't know how to edit the commit message. I can only edit the title "Fix auto-joypad polling"

If you can walk me through it on Discord I'd be most grateful!

bsnes/sfc/cpu/timing.cpp
Screwtapello: That was very well explained, thank you!

Screwtapello:
I'll leave this open for a little while in case @jbo-85 wants to take a look, but I'm pretty happy with this.

Screwtapello (approved):

GrieverV:
You'll also probably want to remove the hacks from [bsnes/target-libretro/program.cpp](https://github.com/bsnes-emu/bsnes/blob/d0efbfbcd937142969b47d65a04ffe4f5a77b50d/bsnes/target-libretro/program.cpp#L164).

rtretiakov:
@asuramaru great work!

asuramaru:
> You'll also probably want to remove the hacks from [bsnes/target-libretro/program.cpp](https://github.com/bsnes-emu/bsnes/blob/d0efbfbcd937142969b47d65a04ffe4f5a77b50d/bsnes/target-libretro/program.cpp#L164).

Thankfully the way the hacks work is string-based so the result is they just get ignored, but it's a good idea to clean up that file as well, good catch!

Screwtapello:
Since removing the hacks from the libretro target was simple enough, I made the change myself. I intend to merge this branch once all the builds have succeeded.

PR #82, opened 2020-10-27
nall: fix string.format() bug reaching 1 byte beyond allocated array bounds

JamesDunne:


Alcaro:
I can reproduce a Valgrind warning in string::format using
```
#include <nall/string.hpp>
int main() { puts(nall::string("{0}").format("5")); }
```
but your proposed patch does not fix it.

From what I can see, the correct fix would be `memory::move(&data[x], &data[x + difference], remaining - difference);`.

JamesDunne:
Ah that might be better. I didn't do any comprehensive testing myself.

JamesDunne:
Trying to replicate the valgrind issue on my mac but valgrind is having compatibility issues with anything newer than High Sierra. On Mojave here.

JamesDunne:
Finally was able to reproduce the valgrind issue on linux and patched appropriately.

Screwtapello:
Thank you very much @JamesDunne for spottiing that error, and thank you also to @Alcaro for reviewing it more quickly and thoroughly than I could!

PR #89, opened 2020-11-02
target-bsnes: Patch up broken recent-game entries.

Screwtapello:
In f09c45f, we fixed native file picker dialogs, so they would correctly set the "option" field to Auto when loading games, instead of leaving it set to an empty string which effectively forced NTSC mode. However, this wouldn't fix up the old entries in the Recent Games menu, so we also need to patch those entries when we read them.


Marat-Tanalin:
This approach results in duplicate entries in the recent-games’ list.

```
Game
  Recent
    1: Auto;C:/example.zip
    2: ;C:/example.zip
```

Screwtapello:
Huh, so it does. I had hoped I was late enough in the process that I'd just be transparently fixing the old entry, not creating a new entry along the way.

Screwtapello:
Closing in favour of #92

PR #91, opened 2020-11-03
Add some documentation about what's in various bsnes directories.

Screwtapello:
This is of course not nearly complete, but it's more than we had previously, and might inspire people to submit their own observations.

### DIFF
--- a/bsnes/README.md
+++ b/bsnes/README.md
@@ -1,0 +1,53 @@
+bsnes source code
+=================
+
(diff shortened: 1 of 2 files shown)


Screwtapello:
You can see what the formattted version looks like [in my repository](https://github.com/Screwtapello/bsnes/tree/74a1f3f3f494b9d92756160476266cb9693fc492/bsnes).

Issue #51, opened 2020-07-31
Verify bsnes supports loading external manifests instead of heuristic manifests.


XXLuigiMario:
The December 1994 build of Star Fox 2 (also known as the Winter CES build) black screens on the latest version of bsnes.
Music for the intro will play without displaying any graphics until it eventually gets stuck playing the same sequence of notes over and over.

`sha1: 78bd997be4cf54b952dec88957d59e8ec30c7723` (SF.ROM)

The header is misplaced, bytes 7FB0-7FBF need to be copied over to 7FC0-7FDF. Tested working on Snes9x.

Screwtapello:
Hm. No-Intro [lists four betas for Star Fox 2](https://datomatic.no-intro.org/index.php?page=search&op=datset&sel_s=Nintendo+-+Super+Nintendo+Entertainment+System&stext=Star+Fox+2&where=1&searchme=Search&pageSel=0&element=Titles&sort=Title&order=Ascending), from 1994-04 to 1994-05 and 1995-06. None of those seem likely to be the build used at Winter CES in January '95.

The SHA1 you listed does not appear in the current No-Intro set, even as a bad dump for one of the other betas, and Google has never heard of it.

XXLuigiMario:
@Screwtapello Doesn't look like it has been datted into No-Intro yet. It was found in the July 24th leak.
It does however have a [TCRF page](https://tcrf.net/Proto:Star_Fox_2/December_28,_1994_(CES)). The hash matches the listing [here](https://github.com/XXLuigiMario/gigaleak/blob/master/other.7z/SFC.7z/SFC/%E3%82%BD%E3%83%BC%E3%82%B9%E3%83%87%E3%83%BC%E3%82%BF/StarFox2/SF2/SF2_CES.ZIP/SF.ROM).

Screwtapello:
Ah, right, thanks, that explains things.

It's OK and expected for betas, protos and unreleased games to not work correctly in bsnes unless the internal header is hacked to trick the emulator into emulating a compatible memory-map, or unless a manifest is provided. I don't count this as a bug in bsnes.

However, while I trust external manifests work properly in higan, I'm not 100% certain they work correctly in bsnes. I'm going to leave this issue open to remind me to verify that bsnes can load external manifests, and to document exactly how that works.

Screwtapello:
I looked into it, and yes, bsnes supports loading external manifests.

[...]

PR #93, opened 2020-11-04
Add a skeleton for bsnes user documentation.

Screwtapello:
As I was doing the research for #51, I figured I should write down what I learned before I forgot it, and if I was going to do that, I might as well write it down in Markdown so we can make a website from it.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,5 @@
+# bsnes User Documentation
+
+bsnes is a multi-platform Super Nintendo (Super Famicom) emulator, originally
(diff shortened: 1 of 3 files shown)


Screwtapello:
The resulting website is not (yet) online, although I intend to use ReadTheDocs to host, like I did for the higan documentation. You can read the actual first article on GitHub here: https://github.com/Screwtapello/bsnes/blob/2ac20577feb24e58265a1d761d41bbf9dda41885/docs/how-to/manifest.md

Issue #75, opened 2020-10-19
Lethal Enforcers does not register shots


rtretiakov:
This could be two separate issues, but I suspect they are related.

When playing Lethal Enforcers, it seems impossible to get past the target "calibration" screen. It's almost as if something is "stuck". It could be that I am simply doing something wrong, but when I tried implementing this feature in libretro I could not get it working there either.

Another thing to add is that when setting up input, simply clicking the mouse twice when trying to assign the trigger will cause a segmentation fault - in other words, you must click the options at the bottom of the window for one of the three mouse buttons, but nowhere else in the window or you will segfault. This is the case when there is a game playing, or when there is no game playing. It also does not seem to be affected by what the controller port 2 device is set to in the main window.

OS: Linux (maybe others?)

Screwtapello:
I tried to set up a Justifier, clicked about, and crashed bsnes. I rebuilt in debug mode, tried again, and couldn't reproduce the crash. Then I rebuilt *outside* of debug mode, and still couldn't reproduce it. I don't doubt there's crashes in that code, I'm sure it's been ticklish before, but I can't reproduce it now.

I can also confirm that Lethal Enforcers doesn't seem to work.

- Configure the bindings for the Justifier in port 2, mapping X and Y to the mouse axes and Trigger and Start to the buttons
- Configure the "Toggle Mouse Capture" hotkey
- Connect the Justifier to port 2
- Load up *Lethal Enforcers*
- I think if it detects a Justifier connected, it takes you straight to the "calibration" screen, with a target, a blinking "shoot at screen" message, and messages in the top left and top right saying "aim at screen" for player 1 and player 2
- There's a green crosshair cursor on the screen, if you press the Toggle Mouse Capture button you can control it more reliably
- Pressing the Trigger binding makes the screen flash white
- The game never seems to accept that as valid input, and the messages in the top-left and top-right never change.

Screwtapello:
"De andere Max" on Discord figured out the crash occurs when you click below the last item in the table, while assigning an input. From that it was easy enough to get a backtrace and find the problem. Thanks for reporting it!

Screwtapello:
I updated my previous comment to include proper verified steps to reproduce.

[...]

PR #95, opened 2020-11-05
target-bsnes: Fix crash when clicking in an empty row while assiging inputs

Screwtapello:
When assigning an input, you can click on a different row of the bindings table to change which input you're assigning. As well as listening for keyboard and controller input, bsnes displays buttons to map the mouse axes or mouse buttons, depending on whether the new row you've selected represents an analog or digital input.

However, some controllers do not have enough inputs to fill the default size of the input table. If you click below the last row, then *no* input is selected, and when bsnes tries to find the input associated with that not-a-row, it crashes.

The solution is, only show the mouse axis/mouse button buttons when a valid row is selected.

Mentioned in #75.

### DIFF
--- a/bsnes/target-bsnes/settings/input.cpp
+++ b/bsnes/target-bsnes/settings/input.cpp
@@ -67,7 +67,7 @@ auto InputSettings::updateControls() -> void {
   assignMouse2.setVisible(false);
   assignMouse3.setVisible(false);
 
-  if(activeMapping) {
+  if(activeMapping && batched) {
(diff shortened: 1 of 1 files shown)


jchv:
I mentioned it in Discord, but this fix is incorrect. The batched state should only be used for, well, batch operations, like clear and assign. Using the first index of the batch state here is presumably a mistake and instead activeMapping should be used. Otherwise, the UI state will be updated as if you were assigning to a different input than you actually are!

Screwtapello:
Closing in favour of #98.

PR #96, opened 2020-11-06
Fix EXHIROM-RAM-* board mappings.

jchv:
It is unclear why these addresses were added to the mapping. Removing these two mappings brings behavior inline with SNES9x:

https://github.com/snes9xgit/snes9x/blob/master/memmap.cpp#L2812

Corresponding Higan change: https://github.com/higan-emu/higan/pull/135


jchv:
Context from @byuu_san on Twitter: https://twitter.com/byuu_san/status/1324479722735087617

> ... why did I have to go and do this. That is so blatantly wrong and knocks out 896KB of my expanded translation space. I am going to have to ship it with manifests or hope they can get a new release out with that line removed.

>Specifically that maps RAM on top of the lower 32KB of for the last 896KB of ROM. So technically I lose 448KB of ROM but I kind of needed the 64K bank ranges there since DMA can't wrap banks on the SNES. This only affects >56mbit ROMs, of which there are none currently.

Screwtapello:
Thank you (and @asuramaru!) for jumping on this and figuring it all out, so I could wake up and pretty much just hit merge!

PR #98, opened 2020-11-07
target-bsnes: Fixes for a couple of input settings bugs.

jchv:
-   Fixes a bug where the program may crash or display improper buttons when changing the selected row during assignment.

    This appears to just be a mistake, accidentally using the 'batched' rows instead of the activeMapping row.

-   Fixes a bug where starting multiple assignments without finishing one leads to stale "Assigning..." text in some rows.

    This appears to be an oversight. Simply refreshing the mappings before starting an assignment fixes this.

This patch should supersede #95 if testing proves that it works as intended across all platforms as it is more straightforwardly-correct (this fixes not only a crash, but also a logical error that can lead to at least one impossible assignment.)

| Before | After |
| --- | --- |
| Animation displaying several input bugs as described above. | Animation showing the inability to reproduce the bugs described above. |

### DIFF
--- a/bsnes/target-bsnes/settings/input.cpp
+++ b/bsnes/target-bsnes/settings/input.cpp
@@ -68,12 +68,11 @@ auto InputSettings::updateControls() -> void {
   assignMouse3.setVisible(false);
 
   if(activeMapping) {
-    auto& input = activeDevice().mappings[batched.left().offset()];
-    if(input.isDigital()) {
+    if(activeMapping->isDigital()) {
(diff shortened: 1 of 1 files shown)


Screwtapello:
Thank you for double-checking my attempted fix, and thank you for so comprehensively describing your PR.

I've linked to CI binaries built from this branch on Discord, and hopefully people will test them out. If nobody finds any problems (because everything worked, or because nobody bothered to test it) in a day or two, I intend to merge this.

PR #99, opened 2020-11-08
target-bsnes: Rework screensaver suppression

### LINKED ISSUE #87: Disable screensaver while playing on macOS
Hi 😊
I'm using the current nightly on macOS Catalina.
how do I prevent screensaver (or monitor /computer sleep) from messing up with game play?
Thanks
-a-

Sintendo:
Fixes #87.

This PR implements the missing hiro/cocoa functionality to make this feature work on macOS and makes the setting configurable in the UI. Additionally, because I felt it doesn't make sense to suppress screensavers when nothing is loaded, I made some changes so the feature only kicks in when emulation is active.

### DIFF
--- a/bsnes/target-bsnes/bsnes.cpp
+++ b/bsnes/target-bsnes/bsnes.cpp
@@ -40,7 +40,6 @@ auto nall::main(Arguments arguments) -> void {
 
   settings.load();
   Application::setName("bsnes");
-  Application::setScreenSaver(settings.general.screenSaver);
   Application::setToolTips(settings.general.toolTips);
 
   Instances::presentation.construct();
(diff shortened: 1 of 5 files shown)


Screwtapello:
Thank you very much for working on this - for the macOS support, and also for working on screensaver stuff which has been... unreliable on X11 at least.

I can't test the macOS changes, but I can confirm that under XFCE, my screensaver kicks in when bsnes is not running, when bsnes is running without a game loaded, and when bsnes has a game loaded but paused, while my screensaver does *not* activate while bsnes is actively emulating a game.

I'll try under GNOME next.

Screwtapello:
OK, it turns out it *doesn't* work under GNOME, but for unrelated (and long-term) reasons that shouldn't hold up this PR. Again, thank you!

PR #103, opened 2020-11-09
Use XDG base directories

ghisvail:
When available on UNIX / Linux systems

### DIFF
--- a/nall/path.hpp
+++ b/nall/path.hpp
@@ -90,7 +90,12 @@ inline auto userSettings() -> string {
   #elif defined(PLATFORM_MACOS)
   string result = {Path::user(), "Library/Application Support/"};
   #else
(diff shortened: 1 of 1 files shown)


Screwtapello:
Thank you for doing this! I've always wanted nall to be more XDG-compliant, but I never quite got around to it. Hopefully someday nall will learn to search through all the XDG fallback directories instead of just grabbing the first thing it finds... but this is definitely a good start.

Also, thank you for submitting PRs to the higan repo as well - I've been head-down working on triaging bsnes bugs and reviewing PRs, but I've been nearly finished and just about to get stuck into the higan repository for about a month.

ghisvail:
> search through all the XDG fallback directories instead of just grabbing the first thing it finds

Speaking of which, I have submitted #110 as a simple improvement.

PR #104, opened 2020-11-10
Install icons to default theme

ghisvail:
See the [Icon Theme Specification](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html)


Screwtapello:
My recollection is that the bsnes icon was installed that way to avoid having to read and understand the Icon Theme Specification, but if this diff is all we need, then yeah, let's do it!

PR #110, opened 2020-11-10
Add shared data to search paths

ghisvail:
This is the patch I am currently using for the Flatpak to enable lookup for files installed under `/usr/share/bsnes` (read-only, common to all users). User data remains the first directory searched through, with shared data used as a fallback. Considering how file location is implemented in `bsnes`, I believe that's the least invasive approach to support shared data without breaking existing installs.

### DIFF
--- a/bsnes/target-bsnes/bsnes.cpp
+++ b/bsnes/target-bsnes/bsnes.cpp
@@ -12,6 +12,9 @@ auto locate(string name) -> string {
   location = {Path::userData(), "bsnes/", name};
   if(inode::exists(location)) return location;
 
+  location = {Path::sharedData(), "bsnes/", name};
+  if(inode::exists(location)) return location;
+
   directory::create({Path::userSettings(), "bsnes/"});
   return {Path::userSettings(), "bsnes/", name};
 }


ghisvail:
Fyi, that's also the approach Debian took for [higan v106](https://salsa.debian.org/games-team/higan/-/blob/debian/106-1/debian/patches/use_sharedpath.patch)

ghisvail:
> nall's existing behaviour is a compromise between the behaviours of POSIX, Windows, and macOS

It basically provides a platform-independent access to conventional app directories (user data and settings, shared data, temporary directory...). With the merge of #103 and #104, it now has decent support for Linux distros respecting the freedesktop specs.

> the same search is used both for reading data

You are correct, and this mixture of responsibilities may lead to unexpected behaviour.

> where we put data files beside the executable and expect bsnes to only read or write those files, and not touch the per-user or system-wide directories.

System-wide dirs like /usr/share are read-only so I am not too worried. The order of the search path remains the following: the program root dir (useful for pre-built binaries), the user data (useful for per-user content) and the user settings in write mode. Adding the shared data (for user-independent read-only data) between user data and settings sounds ok to me.

> If fixing this correctly is very difficult or complex

You'd be aiming for something like what is done in `higan`, with paths controlled by a settings file instead of computed at runtime for each call to `locate`. That would yield a more invasive set of changes.

> Just that investigation would be more work than this PR

I am happy to keep this patch for the Flatpak until a definitive solution is found.

I'll try to set aside some time to contribute.

Screwtapello:
> System-wide dirs like /usr/share are read-only so I am not too worried.

I recall higan implicitly assumes (or at least, it used to) that updates must be written back to the place the original data was read from; that's why the bsnes and higan makefiles bail out if you try to run `make install` as root. I'd like to be sure that assumption is no longer a problem before letting higan/bsnes read from system-wide directories.

> You'd be aiming for something like what is done in `higan`, with paths controlled by a settings file instead of computed at runtime for each call to `locate`.

That's another whole kettle of fish, and I trust that XDG BaseDir is flexible enough that we can just follow it (slightly modified for the portable-installation use-case) without needing a settings file.

----

I filed #111 for the investigation work I mentioned in my last comment. I'll close this PR for now, but if #111 reveals that this change is a net improvement, we should re-open this and land it.

PR #115, opened 2020-11-20
Fix menu bar background color for GTK3

### LINKED ISSUE #108: hiro gtk3: Black menu bar

ghisvail:
Closes #108

### DIFF
--- a/hiro/gtk/window.cpp
+++ b/hiro/gtk/window.cpp
@@ -158,6 +158,11 @@ auto pWindow::construct() -> void {
 
   gtkMenu = gtk_menu_bar_new();
   gtk_box_pack_start(GTK_BOX(menuContainer), gtkMenu, false, false, 0);
+  #if HIRO_GTK==3
+  GdkRGBA gtkMenuBackgroundColor;
+  gdk_rgba_parse(&gtkMenuBackgroundColor, "lightgray");
+  gtk_widget_override_background_color(gtkMenu, GTK_STATE_FLAG_NORMAL, &gtkMenuBackgroundColor);
+  #endif
 
   formContainer = gtk_fixed_new();
   gtk_box_pack_start(GTK_BOX(menuContainer), formContainer, true, true, 0);


ghisvail:
Not exactly the same gray as GTK2 but close enough!

Capture d’écran de 2020-11-20 17-59-25

asuramaru:
Is there a way to make it match the theme's window color? This could be jarring if using a dark theme. Not on this specific window since the background is painted black, but in general for the toolkit itself.

ghisvail:
> Is there a way to make it match the theme's window color?

Would likely require to use GTK3 in a more idiomatic way, i.e. using CSS styling instead of using deprecated APIs to declare and draw the UI.

asuramaru:
I figure there's gotta be some way to get the current theme window color from GTK, even if it's resorting to something extreme like reading a global setting to get the theme ID and then parsing the main theme resource file to extract it.

ghisvail:
I don't think the app was designed to play nice with themes originally. But
maybe I am wrong. Feel free to figure a better way.

Afaic, this PR addresses the initial issue. I consider better integration
with modern theming practices a separate (and likely more time consuming)
issue.

asuramaru:
I'm not trying to be a thorn in your side. Your patch will break the menubar matching dark themes for me. I don't get the black menubar issue you do. The menubar isn't transparent for me. It seems the transparency is theme-dependent or something.

Here, this works for me and respects my theme window color whether light or dark.

    GtkStyleContext* context = gtk_widget_get_style_context(gtkMenu);
    GdkRGBA gtkMenuBackgroundColor;
    gtk_style_context_lookup_color(context, "bg_color", &gtkMenuBackgroundColor);
    gtk_widget_override_background_color(gtkMenu, GTK_STATE_FLAG_NORMAL, &gtkMenuBackgroundColor);

ghisvail:
It does not work for me.

Adwaita:

Capture d’écran de 2020-11-21 17-29-08

Adwaita-dark:

Capture d’écran de 2020-11-21 17-28-43

Alcaro:
Proposed (highly untested) compromise: Use bg_color if it exists, but if it's transparent, assume it does not exist and use lightgray instead.
```
GtkStyleContext* context = gtk_widget_get_style_context(gtkMenu);
GdkRGBA gtkMenuBackgroundColor;
gtk_style_context_lookup_color(context, "bg_color", &gtkMenuBackgroundColor);
if (gtkMenuBackgroundColor.alpha == 0) gdk_rgba_parse(&gtkMenuBackgroundColor, "lightgray");
gtk_widget_override_background_color(gtkMenu, GTK_STATE_FLAG_NORMAL, &gtkMenuBackgroundColor);
```

ghisvail:
> Use bg_color if it exists, but if it's transparent, assume it does not exist and use lightgray instead.

It always exists apparently, and the alpha value is approx but not exactly zero (~1e-322) for both Adwaita and Adwaita-dark. So the proposed changes yield no improvement, sadly.

ghisvail:
It's gonna be difficult to hack our way out of this, I think. I'll ask again to the GTK support channel.

Meanwhile, if you folks find something that works at least for the default GNOME themes, feel free to submit.

ghisvail:
Actually, there is one, which is the first things I was advised on the GTK support channel : don't impose a black window in the first place.

asuramaru:
> It always exists apparently, and the alpha value is approx but not exactly zero (~1e-322) for both Adwaita and Adwaita-dark. So the proposed changes yield no improvement, sadly.

What about if we modify the alpha we get back to opaque (65535 in GTK parlance)?

> Actually, there is one, which is the first thing I was advised on the GTK support channel : don't impose a black window in the first place.

So I gave this a try and it causes two major problems:

1. the entire window flashes gray before turning black on startup, which might be theme/toolkit dependent.

2. resizing the window leaves ugly gray edges as the viewport struggles to repaint over the newly exposed window edges.

That sounds like typical support channels though. "Why would you want to be able to do a simple thing like get the color of the window? Just redesign your application instead!" 😆

jchv:
Just so it's written somewhere visibly, I personally think that just dropping the black status bar on GTK 3 would not be a big deal.

Perhaps as half-way solution, bsnes could properly invoke dark theme instead, on GTK 3. This would give a black menu bar but the text in it should be legible, yes?

I expect that more toolkits will end up going this direction in the future. It is probably no longer guaranteed that there will be an imperative styling API.

ghisvail:
> "Why would you want to be able to do a simple thing like get the color of the window? Just redesign your application instead!" 

I fail to figure how this snarky comment is moving this issue forward.

>  I personally think that just dropping the black status bar on GTK 3 would not be a big deal.

We were discussing the menu bar earlier. Did you mean that or the bar at the bottom of the window?

> Perhaps as half-way solution, bsnes could properly invoke dark theme instead, on GTK 3.

Do you mean forcing the use of a dark theme on GTK3 (like it's done for some media app like Totem)?

> I expect that more toolkits will end up going this direction in the future. It is probably no longer guaranteed that there will be an imperative styling API.

Modern toolkits are definitely moving away from imperative UI composition in favor of more declarative approaches.

asuramaru:
> I fail to figure how this snarky comment is moving this issue forward.

My condolences, it must be rough going through life without a sense of humor.

I'll keep looking for an actual fix to this problem.

ghisvail:
> forcing the use of a dark theme on GTK3 

Btw, this works for me.

```
+  #if HIRO_GTK==3
+  g_object_set(gtk_settings_get_default(), "gtk-application-prefer-dark-theme", TRUE, NULL);
+  #endif
```

See https://stackoverflow.com/a/16003697

> I'll keep looking for an actual fix to this problem.

Thanks :+1:

Screwtapello:
We get our phones back at lunchtime in jury duty, so I wanted to take a little time to chime in here.

As of right now, I don't think PR is the right solution. The black background on the menu-bar makes the GTK3 port visibly inferior to the GTK2 port, and hardcoding the background colour would make it still visibly inferior, just less so. If we get to a point where GTK2 is no longer widely available, GTK3 would be superior than a broken build, even with a black menu bar, and this patch would be even better. But I think we have a little time yet before we have to make such a decision.

In reference to "just don't set the background of the status bar", I don't think that would help, since the menu is not the same colour as the status bar. I think the menu is the same colour as the video context, and we *do* want to keep that black.

If solving the menu bar requires, or even just becomes easier, if the status bar changes colour -- sure, I'm not married to it.

Switching to a dark theme variant might be a nice idea on every platform that supports it - GTK3 and macOS do, does Windows? I guess there would need to be a hiro API that's ignored on some backends, but that seems like a different, larger issue.

jchv:
Arguably, that hiro API should go away, as it is no longer logical to provide it. It makes sense for win32 and gtk2, but I think it makes less and less sense as time goes on.

Screwtapello:
When you say "that hiro API", which API are you talking about?

ghisvail:
> If we get to a point where GTK2 is no longer widely available

It's no secret that the GTK developers and Linux maintainers want GTK2 gone. GTK3 is in maintenance mode and GTK4 is where the action is happening. The last release of GTK2 was from 2011. Should a vulnerability be found, it won't be fixed.

Besides, although GTK2 may still be around for a little while, other dependencies may not. For instance, Debian got rid of gtsourceview2 for the next stable release.

> Switching to a dark theme variant might be a nice idea on every platform that supports it. GTK3 and macOS do, does Windows?

The Adwaita themes, including the dark variant, are installed on all platforms. If someone could test it for Windows, I can submit a different PR with the dark theme preference set.

> When you say "that hiro API", which API are you talking about?

I guess @jchv is talking about the code path that composes the UI in hiro. Some of it is written in an imperative style which might prove difficult to twist to support modern toolkits effectively.

asuramaru:
I made a more comprehensive patch, I hope it will be to everyone's liking.

https://github.com/bsnes-emu/bsnes/pull/116

Screwtape, jury duty comes first, please don't feel any obligation to get to this until you're back and well rested.

> For instance, Debian got rid of gtsourceview2 for the next stable release.

This is only used by hiro::SourceEdit, so we can disable it in the hiro/components file when Debian 11 ships. That won't buy us much more time but it's something at least.

ghisvail:
Closing this PR in favor of #116 or #118

Kawa-oneechan:
> Switching to a dark theme variant might be a nice idea on every platform that supports it - GTK3 and macOS do, does Windows?
Win32 does not. You can detect if Win10 is set to dark mode via a registry key (which strangely says if it's set to *light* mode) and if it's missing nothing needs break. Flip side, not all Win32 controls, *such as the menu bar*, support having their colors changed in a consistent manner. Some controls respond well enough to the "set background color" message, like text fields and push buttons, but checkboxes (a kind of button ffs!) do not and require you to actually draw them by hand. That's how classic VB managed to support back/forecolor properties for checkboxes and how WinForms still does. I can only assume this holds for menus.

Source: I implemented darkmode in Clunibus, a Win32 app.

PR #116, opened 2020-11-24
Fix hiro GTK3 menubar color when using custom window background color (fixes #108)

### LINKED ISSUE #108: hiro gtk3: Black menu bar

asuramaru:
When bsnes (or other hiro application) set a custom background color,
the GTK3 menubar color takes on the custom window background color.
So for instance in bsnes, this makes the menubar back, with black text,
making the menubar unreadable.

What this patch does is use the default menubar colors when no custom window
background color is used. When one is set (or removed) via
Window::setBackgroundColor, the new pWindow::_setMenuColor function is called.

This function sets the default menubar colors if the custom window color
option has been disabled, and sets it to the GTK3 theme background color,
which is the same as the menubar color, if a custom window color has been set.
We don't have to modify the text because there is no hiro option to set the
parent window text color.

If retrieving the theme background color value fails, and we have a custom
window color, this patch has one last fallback where it will set a light gray
menu color with black text.

On GTK2, this patch has no effect as it is not needed. GTK2 doesn't use
transparent menubars.

### DIFF
--- a/hiro/gtk/window.cpp
+++ b/hiro/gtk/window.cpp
@@ -158,6 +158,7 @@ auto pWindow::construct() -> void {
 
   gtkMenu = gtk_menu_bar_new();
   gtk_box_pack_start(GTK_BOX(menuContainer), gtkMenu, false, false, 0);
+  _setMenuColor();
 
   formContainer = gtk_fixed_new();
   gtk_box_pack_start(GTK_BOX(menuContainer), formContainer, true, true, 0);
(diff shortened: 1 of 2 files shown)


asuramaru:
This fix was tested with multiple Gnome 3 and Xfce 4 themes under the Gnome 3 and Xfce 4 desktop environments.

Screenshot_2020-11-24_04-22-09

asuramaru:
@ghisvail merged my patch into his, I'm happy with that.

PR #117, opened 2020-11-24
Prefer dark theme on GTK 3

ghisvail:
Following discussion in #115, could be complementary to #116.

### DIFF
--- a/hiro/gtk/window.cpp
+++ b/hiro/gtk/window.cpp
@@ -207,6 +207,10 @@ auto pWindow::construct() -> void {
   g_object_set_data(G_OBJECT(widget), "hiro::window", (gpointer)this);
   g_object_set_data(G_OBJECT(formContainer), "hiro::window", (gpointer)this);
 
+  #if HIRO_GTK==3
+  g_object_set(gtk_settings_get_default(), "gtk-application-prefer-dark-theme", TRUE, NULL);
+  #endif
+
   pApplication::state().windows.append(this);
 }
 


asuramaru:
hiro is used by more than just bsnes-emu. This patch would make every hiro application run in the dark theme mode including higan-emu, icarus, and any third-party applications that try and make use of hiro.

hiro is a library meant to be used by anyone. This should be implemented as an API, such as Application::setPreferredTheme(Theme::Dark) that bsnes-emu then calls during startup, similar to Application::setName.

ghisvail:
> This patch would make every hiro application run in the dark theme mode including higan-emu, icarus, and any third-party applications that try and make use of hiro.

Indeed, this is a good point.

The purpose of this PR was to move the focus of the "dark theme as default" to a separate discussion.

> Application::setPreferredTheme(Theme::Dark) that bsnes-emu then calls during startup, similar to Application::setName

I like this as a solution to your point above.

> You might want to consider if it's the HID convention for KDE, Windows and macOS next

Using your proposal, this logic could be restricted to just Linux using appropriate guards, and later expanded to macOS and Windows if supported?

asuramaru:
> Using your proposal, this logic could be restricted to just Linux using appropriate guards, and later expanded to macOS and Windows if supported?

Yes, there's two ways we could do this.

One is that Application::setPreferredTheme or setPrefersDarkTheme or whatever we call it will set a flag inside Application::state itself. GTK3 will simply read out that state and then set the relevant property. The other OSes will do nothing until and unless it's implemented later on.

The other is that we make the Application::setPreferredTheme call into pApplication::setPreferred theme, and we implement dummy stubs (empty functions) in the Windows, Cocoa and Qt ports. If we do this, we don't have to add a new state variable to Application::state.

I prefer the first option but it doesn't really matter.

I guess we should bikeshed the API name now. Screwtape, jchv, any preferences?

1. Application::setPreferredTheme(Theme = Theme::Light | Theme::Dark)
2. Application::setPrefersDarkTheme(bool = true)
3. something else?

> The purpose of this PR was to move the focus of the "dark theme as default" to a separate discussion.

That's fair. It sounds like whether I prefer it or not, it's the HID rules for Gnome so it probably should be done for GTK3. I tend to use Clearlooks which doesn't have a dark theme anyway so it shouldn't affect me.

Screwtapello:
Again, sorry it's taken me so long to review this PR.

hiro is *supposed* to be a standalone library suitable for all kinds of applications, but in practice higan and bsnes are by far the most popular apps built on it, and they're both emulators, so it would make sense for them both to use dark mode on platforms that support it. So, I could be talked into merging this change as-is.

However, I agree with asura that making it an API would be nicer than hard-coding it. The exact implementation would depend on the platform-specific behaviours we need to abstract over. After some brief searching, it seems:

  - Windows doesn't have a per-application dark theme ([at least, officially](https://github.com/ysc3839/win32-darkmode)), just a system-wide dark theme.
  - Qt doesn't support a per-application dark theme, but will try to respect Windows' system-wide setting
  - macOS lets you switch between light and dark theme [on a per-widget basis](https://developer.apple.com/documentation/appkit/nsappearancecustomization/choosing_a_specific_appearance_for_your_macos_app), using an API similar to GTK+3.

Since the two platforms that support a per-application dark theme both use a procedural API, I think that would be the best model for hiro to adopt:

  - `Application::setPreferDarkTheme(true)` calls `pApplication::setPreferDarkTheme`
  - on GTK+3, we do the `gtk_settings` dance as described in this PR
  - on macOS, we do the `NSAppearance` dance as described in the page I linked
  - on Windows and Qt, we have a do-nothing stub
  - When the situation changes, and Windows and/or Qt do something that makes this API inappropriate, we aggressively refactor it with our new understanding of the problem space

Just setting a flag in the `Application` class and having the GTK+3 backend query it would be a less-invasive change, but I fear that would obscure the link between the code that sets the flag and the code that reads it - somebody might try to set it at runtime, or on a different platform, and wonder why nothing happens. At least if there's a direct call, somebody looking at (say) the Windows backend can see that the function is an empty stub, rather than not finding anything and wondering if the functionality is magically supplied by some other means.

ghisvail:
> I think that would be the best model for hiro to adopt [...]

Sounds good to me.

Screwtapello:
I've filed #136 for the solution I outlined in my previous comment. Since this PR is too narrowly-focussed to solve that issue, I'll close this PR (it's linked from that issue for future reference) and hopefully one of us will get around to it sooner or later.

Again, thank you for your work researching this issue and preparing this PR!

PR #118, opened 2020-11-24
Fix menu bar background on GTK 3

### LINKED ISSUE #108: hiro gtk3: Black menu bar

ghisvail:
Follow-up from #115 with a simpler patch inspired from #116

Closes #108

### DIFF
--- a/hiro/gtk/window.cpp
+++ b/hiro/gtk/window.cpp
@@ -289,10 +289,27 @@ auto pWindow::setBackgroundColor(Color color) -> void {
   gtk_widget_modify_bg(widget, GTK_STATE_NORMAL, color ? &gdkColor : nullptr);
 
   #if HIRO_GTK==3
-  GdkRGBA gtkMenuBackgroundColor;
-  auto context = gtk_widget_get_style_context(gtkMenu);
-  gtk_style_context_lookup_color(context, "theme_bg_color", &gtkMenuBackgroundColor);
-  gtk_widget_override_background_color(gtkMenu, GTK_STATE_FLAG_NORMAL, &gtkMenuBackgroundColor);
+  if(!state().backgroundColor) {
+    // If no window background color is set, then use the default menu bar color.
+    gtk_widget_override_color(gtkMenu, GTK_STATE_FLAG_NORMAL, nullptr);
+    gtk_widget_override_background_color(gtkMenu, GTK_STATE_FLAG_NORMAL, nullptr);
+  } else {
+    // Otherwise, fix visual inconsistencies between the menu bar and current theme.
+    GdkRGBA themeBackgroundColor;
+    auto context = gtk_widget_get_style_context(gtkMenu);
+    if(gtk_style_context_lookup_color(context, "theme_bg_color", &themeBackgroundColor)) {
+      // If the current theme defines theme_bg_color, then use it.
+      gtk_widget_override_background_color(gtkMenu, GTK_STATE_FLAG_NORMAL, &themeBackgroundColor);
+    } else {
+      // Otherwise, set the menu bar to black text on a light gray background.
+      // FIXME: Maybe compute these colors based on the window background instead?
+      GdkRGBA textColor, backgroundColor;
+      gdk_rgba_parse(&textColor, "black");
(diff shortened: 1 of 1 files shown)


ghisvail:
Adwaita:

Capture d’écran de 2020-11-25 00-16-31

Adwaita-dark:

Capture d’écran de 2020-11-25 00-16-17

asuramaru:
I made a more thorough patch than this for a reason.

We aren't guaranteed that theme_bg_color will be defined for every theme out there (theme authors can name their colors whatever they want), and there's no reason to manually drive the theme color if the window background color isn't set. My patch won't cause any difference in behavior if the window background color isn't changed, which is the case for higan-emu and icarus. And if there is a custom background color and the theme_bg_color property is missing, a reasonable fallback will be set with my patch. With your patch, every hiro application would break, drawing a black background with black text, if theme_bg_color were missing.

ghisvail:
I am not quite sure why anything needs to be done outside of `setBackgroundColor`. Afaik, the issue is that the app sets a black background with a transparent menu bar on GTK 3 instead of solid on GTK2, unless overridden by some custom GTK3 theme which we have got no control of. Hence why I targeted the default GTK themes (Adwaita and Adwaita-dark) as representative of well-behaved themes.

Apart from a possibly missing guard on the alpha value, what else is missing in this patch according to you?

> I'm sorry you've been put in this spot, I was just trying to help.

There is nothing to feel sorry about. We both care enough to propose things to improve a piece of software we love on our spare time. That's a luxury not all opensource projects have. I also take this opportunity to thank you for your patience in filling the gap of knowledge we collectively may be missing regarding GTK 3.

asuramaru:
> I am not quite sure why anything needs to be done outside of setBackgroundColor

You're right, I could move the _setMenuColor code inside setBackgroundColor, since construct calls setBackgroundColor later on anyway. I was trying to follow the hiro convention with the other _setMenu functions, but it would reduce the patch size by a couple lines and save an extra function.

> Apart from a possibly missing guard on the alpha value, what else is missing in this patch according to you?

Handling themes that don't use Adwaita's theme_bg_color naming scheme. The names of colors are up to the theme to decide and are not mandated by GTK at all, so I wanted to have a fallback in case it's missing or is called something else in another theme.

I also think we should check if the background color value isn't set (if(!state().backgroundColor) does the job), and not bother trying to override the BG color in that case. Imagine a theme that uses a separate menu-bar color in its theme, we would be overriding that for no reason.

If we take lines 496-516 of my patch and put it inside setBackgroundColor, I think that would do the job. It's only 20 lines vs 5.

https://github.com/bsnes-emu/bsnes/pull/116/files#diff-0a98c445cff13c912408aa7b65814cddd09eeac815a4709ce5c63ff8f6fc6d43R496

I don't want to bikeshed too much, but I also thought about an alternate fallback where if theme_bg_color isn't there, we set the menubar to the window background color, compute the luminance of the background color and choose white or black text based on whichever has more contrast (it's very simple, 3-4 lines of code). That might be better than falling back to light gray always. But maybe we should wait until we find a theme without theme_bg_color to care?

> There is nothing to feel sorry about.

I just don't want Screwtape to have to choose a patch to reject and possibly feel bad about it.

Thank you in any case, sorry we somehow got off on a bad foot. It's great that you're helping too, bsnes/higan has always needed distro patches and it's good that as a community project we can make it more conformant to modern standards like XDG.

ghisvail:
@asuramaru I merged your patch with mine and added you as co-author. Feel free to make another pass at it to check whether I forgot anything. I have also added a few comments to explain the flow of the GTK 3 code path, and left a FIXME section for you where we could try to compute the text and background colors off the window background as you suggested.

asuramaru:
lgtm, thank you ❤️
I closed my request then. I'm not worried about credit in any case. Not until the VC offers start pouring in at least 😆

Screwtapello:
Thank you to both @asuramaru  and @ghisvail for your work on this PR and the research that went into it. However, I've decided to merge #127 instead, since that seems to be a more robust solution.

I hope you'll consider contributing again in future!

ghisvail:
@Screwtapello Glad to see you reached a solution you're comfortable with and that our research helped :+1:

Issue #70, opened 2020-10-11
Support GNU Autotools-style $prefix and $DESTDIR build settings


Screwtapello:
Makefiles produced by GNU Autotools support two separate build settings that control where `make install` will install files:

  - `$prefix` defaults to `/usr/local`, describes where files will be at runtime, and may be compiled into the executable
  - `$DESTDIR` defaults to the empty string, describes where files will be placed at install time, and must not be compiled into the executable

The idea is that when building an RPM or DPkg package, distro maintainers can set `$prefix` to the path where the program will be installed on the target system, and set `$DESTDIR` to build it in some isolated temporary location that doesn't need root access to write to. For example, if the makefile says:

    install out/myprog $(DESTDIR)$(prefix)/bin/myprog

...then a user may run `make install` to get `/usr/local/bin/myprog`, but a distro package may run `make DESTDIR=/tmp/build/myprog prefix=/usr install` to get `/tmp/build/myprog/usr/bin/myprog`, which gets archived into a package, which gets installed as `/usr/bin/myprog`.

bsnes does not support this scheme. There are no build settings to control what paths get compiled into the executable, so the only thing that matters is where files are placed at install time, which is effectively a combination of `$prefix` and `$DESTDIR`. In bsnes' build system, this setting is confusingly named `$prefix`.

However, while bsnes system is simple and self-consistent, it's inconsistent with the rest of the world and causes confusion and grief for people trying to integrate bsnes with any kind of larger packaging system. Although it makes bsnes a little more complicated, we should add `$DESTDIR` support to the build system, probably along the lines of:

- add a new, empty `DESTDIR` variable to `nall/GNUmakefile`, near where `prefix` is declared
- add `$(DESTDIR)` to all the Linux/BSD install commands in `bsnes/target-bsnes/GNUmakefile`

Screwtapello:
Note that supporting *all* the GNU Autotools install path variables probably isn't going to happen, because that's a lot of work and if we really wanted to be compatible with GNU Autotools, we'd just use it as our build system directly.

Issue #111, opened 2020-11-11
Investigate XDG BaseDir integration


Screwtapello:
Currently, bsnes more-or-less supports [the XDG BaseDir specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for storing configuration on Linux - it keeps some things in `~/.config/bsnes/` and some in `~/.local/share/bsnes/` and since #103 landed it even respects `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME`.

However, bsnes' external-resource handling is really designed for the Windows model, where resources tend to be next to the executable, or in the user's home directory, and the distinction between config/data/cache and the user/system/vendor hierarchy don't really exist. I recall that bsnes and higan have at different times put everything in `~/.config` or put everything in `~/.local/share` for simplicity's sake, and I don't know if they're any more rigorous now.

Updating bsnes to properly follow the XDG BaseDir spec on Linux without regressing Windows or macOS may be difficult, but I'd like to get an idea for *how* difficult it would be:

- [ ] Investigate all the places where bsnes loads an external resource
    - a good start would be finding all the things that call functions in `nall/path.hpp` and tracing their callers
- [ ] figure out their access (read/write), and their type (config/data/cache)
- [ ] figure out a search algorithm for each combination of categories (read config, write config, read data, ...)
- [ ] check whether nall provides all the ingredients for discovering the paths required by the search algorithm (I *think* the raw parts are in `nall/path.hpp` and the path construction operations are in `nall/location.hpp` but it would be good to check)

Once we have a good idea of where we are, and where we want to be, we can estimate how much work it would be to get from one to the other, and whether that investment is worth making.

Screwtapello:
I expect the outcome of this issue to be a new issue ("Implement XDG BaseDir integration") that describes exactly the work to be done. If it turns out to be a lot of work, and something easier like #110 is a net improvement, we might just go for that instead.

ghisvail:
Subscribed. Thank you for this very good summary!

Regarding #110, another approach could be to just copy the default stuff installed under `/usr/share/bsnes` to their relevant location under `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME` and not override the search paths in `locate`. This could be easily done with a wrapper script doing this check first, copy if data is absent, and finally execute the binary. Since both locations are r/w, it should be fine.

Screwtapello:
[...]

Issue #100, opened 2020-11-08
Flatpak build


ghisvail:
Same motivation as for [higan](https://github.com/higan-emu/higan/issues/142).

PR #131, opened 2020-12-28
Improve 'make install' for distros.

orbea:
Fixes https://github.com/bsnes-emu/bsnes/issues/70.

The main goal of this is a more standard `make install` and `make uninstall` which will work much better with distros. I only touched the linux/bsd path and left both windows and osx alone since I am not knowledgeable of what those platforms expect.

More specifically this does:

1. Removes the check disabling the `install` and `uninstall` checks when run as root. Some distros like Slackware build and create packages as root and there are tools like sandbox (https://wiki.gentoo.org/wiki/Project:Sandbox) designed to catch programs that install out of the expected paths.
2. This adds `DESTDIR` which by default is undefined. When defined it allows for staged installs. For example everything gets installed into a fake root directory which is then used to create a distro package.
3. This follows the FHS defaults for installed paths where applicable.
4. This unhardcodes `/usr/share/` when `DATADIR` is defined during the build so that it correctly respects the installed path.

For `4` the default paths are:
* prefix = /usr/local
* bindir = $(prefix)/bin
* datarootdir = $(prefix)/share
* datadir = $(datarootdir)

These can be overridden with setting these variables as make arguments.

For example on Slackware I would use:
```
make install \
  prefix=/usr \
  bindir=/usr/games \
  datadir=/usr/share/games \
  DESTDIR=$PKG
```
It is important to have both `datarootdir` and `datadir`as the former is for files like icons and desktop files while the latter is data files specific for bsnes. On Slackware I would install the former to `/usr/share/` and the latter gets installed to `/usr/share/games/bsnes`.

Please review to make sure there are not any hardcoded paths I missed.

Some things I did not include.
* Out of tree builds, it is helpful to be able to compile bsnes without touching the actual source tree. I will look into this later depending on the outcome of this PR.
* Variables like `prefix` or `bindir` may be more standard if written as `PREFIX` or `BINDIR`, but its not really important.


ghisvail:
That would also benefit the [Flatpak packaging](#100).

orbea:
I added one more commit to take care of out of tree builds.

Example:
```
mkdir -p /tmp/build
cd /tmp/build
make install -f /path/to/bsnes/bsnes/GNUmakefile
```

This adds dependencies for the `obj` and `out` directories which may not exist in the build directory so they are automatically created before any compilation.

It may need to abstract `mkdir` similar to `delete` for `rm`, but someone else will need to tell me the proper windows command for this to work.

Screwtapello:
Sorry it's taken such a long time to get back to you!

Most of these changes look pretty straightforward, and in my limited experience trying to write RPM spec files at a former job, I understand why they're useful. However, I'm less sure about supporting out-of-tree builds — that's the biggest change of all (67 insertions, 55 deletions) and I don't think it's strictly needed - I can imagine it being useful in occasional, niche situations, but I don't think it's going to be used regularly enough to make the change worthwhile.

I'm happy to merge the other changes as-is right now, but if you have a common use-case for out-of-tree builds I'm willing to hear it before I decide.

ghisvail:
I believe out-of-tree builds are an orthogonal concern to the original topic of this PR, i.e. proper support for a `make install` target. 

Perhaps it is best to merge the changes until the last commit, submit the latter as a separate PR and continue the discussion there?

orbea:
I removed the out of tree commit from this PR and pushed it to this branch for now

https://github.com/orbea/bsnes/tree/out_of_tree

> However, I'm less sure about supporting out-of-tree builds — that's the biggest change of all (67 insertions, 55 deletions) and I don't think it's strictly needed - I can imagine it being useful in occasional, niche situations, but I don't think it's going to be used regularly enough to make the change worthwhile.

This is a very understandable concern and when I made similar changes for other projects (i.e. mupen64plus) it was not nearly as big of a diff. This is because each file has its own make rule which is a strange design. It would be much better if they are listed in variables with a few make rules to control them all. Perhaps if this was done first it would be a far more straight forward and less intrusive commit.

> but if you have a common use-case for out-of-tree builds I'm willing to hear it before I decide.

Its very useful for source based distros like what is used in gentoo as it keeps a clean distinction from the build files and the actual source tree. It is also helpful in cases where the source directory may not have write permissions or if the user wishes to keep the source on their filesystem while only building in tmpfs.

I also have a personal simple build repo which depends on out of tree builds working so I have a strong desire for it.

https://notabug.org/orbea/build/src/master/games/bsnes

orbea:
I think the freebsd build issue was introduced after I rebased to master...

ghisvail:
@orbea you might want to rebase onto `master` to benefit from this [update](be0ee7652ca2860bf4094a05cdf0b456e1e7d16c).

Screwtapello:
Thank you very much for making that change. I've reviewed this PR, I like everything in it, and I'd be happy to merge it except that I've just noticed it wouldn't work. When bsnes looks for a resource — the game database, shaders, firmware files, etc. — it calls a function called `locate()`:

https://github.com/bsnes-emu/bsnes/blob/be0ee7652ca2860bf4094a05cdf0b456e1e7d16c/bsnes/target-bsnes/bsnes.cpp#L8-L17

It checks beside the executable, it checks the "user data" path (`~/.local/share/bsnes` on POSIX), and then it assumes the file is in the "user config" path (`~/.config/bsnes` on POSIX). It doesn't fallback to `/usr/local/`, it doesn't respect `$DATADIR`, and it definitely doesn't respect `$XDG_DATA_DIRS`.

Probably the simplest thing to do now would be to add a fallback to `DATADIR` or `/usr/local` to that function, but probably at some point we should clean up all bsnes' resource-loading and make it more XDG-compliant. Heck, I might as well file an issue for that now.

Screwtapello:
Somehow I overlooked that such an issue already exists: #111.

orbea:
@Screwtapello Thanks for pointing that out, I failed to notice that `sharedData()` is not actually used. I updated the last commit, I think it should be better now?

ghisvail:
@Screwtapello indeed, that's what we observed back then with my FHS compliance PR on Higan which then triggered issue #111 for further discussion. I did a (hopefully) comprehensive review of path queries throughout the code base and the results are available [here](https://github.com/bsnes-emu/bsnes/issues/111#issuecomment-731361136).

The last [commit](https://github.com/bsnes-emu/bsnes/pull/131/commits/adafcb4430b6115e4e91e7fe1b5ff3059df1914f) from @orbea is similar to the patch applied to the Debian packaging to support FHS shared data directory.

carmiker:
From my perspective, this PR looks good to be merged.

PR #134, opened 2020-12-30
ruby: Auto detect some dependencies on linux.

orbea:
For source builds its annoying and brittle having to write `sed` commands to automatically remove these from `ruby/GNUmakefile` when the dependencies are not present.

I think its better doing this directly in the build system with `pkg-config --exists`.

This will automatically detect and enable if present:
* sdl2
* openal
* pulseaudio
* libao
* udev

This also only applies to linux builds, other platforms may want to do similar, but I will leave that for people who actually use them.

### DIFF
--- a/ruby/GNUmakefile
+++ b/ruby/GNUmakefile
@@ -1,3 +1,13 @@
+# function pkg_check(pkg_config_name, ruby_name)
+pkg_check = \
+  $(strip \
+    $(if $(shell pkg-config --exists $1 && echo yes), \
+      ruby += $2 \
+    , \
+      $(warning Package $1 not found, disabling driver) \
+    ) \
+  )
+
 ifeq ($(ruby),)
   ifeq ($(platform),windows)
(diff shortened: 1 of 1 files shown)


Screwtapello:
Again, I'm sorry it's taken so long to review this change. I really like this idea, but would it be difficult to print out a message like "Package sdl2 not found, disabling driver" to alert people to the fact that they're missing an optional dependency?

orbea:
> would it be difficult to print out a message like "Package sdl2 not found, disabling driver" to alert people to the fact that they're missing an optional dependency?

It should be possible, I will update this PR with this within the next few days probably.

orbea:
@Screwtapello It will now print warning messages when the pkg-config file is missing, for example.

```
../ruby/GNUmakefile:22: Package PulseAudio not found, disabling driver
```

Screwtapello:
I decided that with the extra error message, it would be nice to factor the check into a helper function, so I made that change and if it passes, I'll merge it. Thank you for the work you put into this PR!

Issue #137, opened 2021-01-08
hiro: do not require gtksourceview-2.0


Screwtapello:
By default, bsnes builds hiro with support for the `SourceEdit` control, which (in GTK+2 builds) is implemented on top of the `gtksourceview-2.0` library. While GTK+2 itself widely used enough to be around for a long time, `gtksourceview-2.0` [has already been removed from Debian Testing](https://tracker.debian.org/news/1173790/gtksourceview2-removed-from-testing/), so removing our dependency on this library is important.

It turns out that bsnes does not actually *use* the `SourceEdit` widget, so it shouldn't be too hard to make hiro build without it, at least for GTK+2 builds. Alternatively, we could just make GTK+3 the default backend on POSIX, with the same effect.

(GTK+3 builds currently depend on the `gtksourceview-3.0` library, which has *also* been deprecated in favour of `gtksourceview-4.0` — still based on GTK+3 — but that's a job for a different issue)

orbea:
I made a PR that disables the dependency everywhere when building bsnes. https://github.com/bsnes-emu/bsnes/pull/139

PR #139, opened 2021-01-08
hiro: Don't depend on gtksourceview when building bsnes.

orbea:
Fixes https://github.com/bsnes-emu/bsnes/issues/137

Seems to all work for me.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,7 +3,7 @@ linux-x86_64-binaries_task:
         image: ubuntu:latest
 
     setup_script:
-        - apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install build-essential libgtk2.0-dev libpulse-dev mesa-common-dev libgtksourceview2.0-dev libcairo2-dev libsdl2-dev libxv-dev libao-dev libopenal-dev libudev-dev zip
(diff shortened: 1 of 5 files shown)


Screwtapello:
Thank you for doing this work, and I'm sorry it's taken a while to get back to you. I hesitated a while over this PR because while it definitely meets the criteria listed in the original issue, I didn't think adding a new Makefile variable just for this dependency was the right way to go.

Now that Near is out of retirement and (at least theoretically) still working on hiro, I asked him how he would solve the problem, since I'd like bsnes (and higan) to easily merge future hiro changes, and he sent me the changes in #163. They don't actually build on macOS yet, so they'll need a little work, but I would like to continue tackling this issue in that PR.

As somebody who's investigated this problem, if you wanted to look over #163 and give feedback, I'd appreciate it — but if you're happy to let it slide, that's fine too, you've already been more than helpful.

orbea:
> As somebody who's investigated this problem, if you wanted to look over #163 and give feedback, I'd appreciate it — but if you're happy to let it slide, that's fine too, you've already been more than helpful.

The approach in this PR was taken as to make it configurable and not require any changes that would affect anything other than bsnes. If you want to just want to entirely remove it that is fine.

However if I am going to be honest, the linked PR is messy. It combines the code changes and CI changes in a single commit, it has unrelated changes and doesn't seem to actually solve the issue...

Issue #148, opened 2021-02-08
Controller input is not recognized under macOS


seeseekey:
Under macOS (Big Sur; 11.2) controller input is not recognized under macOS.

If i try to configure it with the input dialog, the controller inputs are not recognized. In other applications the controller works. Different controllers are tested; same issue.

timotheeg:
I tested my [Buffalo BSGP815GY USB controller](https://www.amazon.com/Buffalo-iBuffalo-Classic-Gamepad-BSGP815GY/dp/B06XWD8QQJ) in 11.2 and had no issue in [bsnes nighly 2021-02-08](https://github.com/bsnes-emu/bsnes/releases/tag/nightly).

What controller model(s) are you using?

seeseekey:
I tested the following controller:

* 8BitDo SFC30 Gamepad
* XBox Wireless Controller
* Stadia Controller

If i try to assign any button of these controllers nothing happens. The controllers are recognized by the OS. Assigment of keyboard keys works.

I use a new Mac mini (2020) with M1 (Apple Silicon) chip.

![bsnes-driversettings](https://user-images.githubusercontent.com/1023818/107474710-202a8280-6b73-11eb-8d6c-6d57dfbb2e1b.png)
![bsnes-assign](https://user-images.githubusercontent.com/1023818/107474713-215baf80-6b73-11eb-8207-bcff4c13cc97.png)
![bsnes-controller](https://user-images.githubusercontent.com/1023818/107474715-21f44600-6b73-11eb-8b4c-079dda96aca4.png)

timotheeg:
Thanks for the extra details. I have a 8BitDo SFC30 Gamepad as well. I just tested (over bluetooth connection), and it is working fine in the same bsnes version, reference screenshot below.

Some differences to your setup however:
* I couldn't test the 8bitdo in 11.2, because I had already done the upgrade to 11.2.1
* I'm on an iMac18,3 (Quad-Core Intel Core i7 - 2017 model)

I don't see anything in the [11.2.1 release notes](https://support.apple.com/en-us/HT211896) update that could affect controller support, but you could try upgrading anyway.

I don't have a mac mini M1 to test further 😞 . Hopefully someone else has a similar setup to help you troubleshoot further.

![image](https://user-images.githubusercontent.com/935223/107523310-8ead2600-6bef-11eb-8518-29825203706f.png)

seeseekey:
I have updated to 11.2.1 today and test again; same issue?

[...]

PR #150, opened 2021-02-11
Support building universal binaries on macOS.

### LINKED ISSUE #149: Support compiling universal macOS binaries.
While Rosetta 2 works fine for most intents and purposes, it seems that gamepads do not work properly under Rosetta 2 applications. (#148)

Relevant information: [Building a universal macOS binary](https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary#3618377).

(Feel free to assign me. I plan on taking this on later tonight.)

jchv:
I've attempted to compartmentalize the added complexity as much as possible. It could be improved also if we could make hiro and ruby use the nall compile function.

I'm going to let this rip on CI to see if it works.

Fixes #149.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 hiro/qt/qt.moc
+.DS_store
--- a/bsnes/GNUmakefile
+++ b/bsnes/GNUmakefile
@@ -60,5 +60,8 @@ include $(ui)/GNUmakefile
 -include obj/*.d
 
 clean:
+ifeq ($(platform),macos)
+	rm -rf out/$(name).app
+endif
 	$(call delete,obj/*)
 	$(call delete,out/*)
--- a/bsnes/obj/.gitignore
+++ b/bsnes/obj/.gitignore
@@ -1,2 +1,2 @@
-*.o
-*.d
+*
+!.gitignore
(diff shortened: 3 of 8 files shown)


Screwtapello:
It seems GitHub Actions does not have a modern-enough macOS toolchain for this, yet: https://github.com/actions/virtual-environments/issues/2486

It would be nice to teach nall's `$(compile)` function to accept the extra flags that hiro and ruby need, instead of those libraries defining their own version.

I also wonder if the "universal binary" flag option should be `local=false`, since "only compile for my CPU architecture" is basically the same idea as "only compile for my CPU microarchitecture" but moreso.

seeseekey:
Is there a description how i can compile these branch on macOS (M1) - cd bsnes / make?

jchv:
@seeseekey 
> Is there a description how i can compile these branch on macOS (M1) - cd bsnes / make?

Try this.

```
cd path/to/the/cloned/repo
make -j8 -C bsnes local=false
```

It should work if you have XCode installed. Then there should be a copy of bsnes in `bsnes/out/bsnes.app`.

I did also provide an M1 build over at https://github.com/bsnes-emu/bsnes/issues/148#issuecomment-777009602 in case you missed it.

seeseekey:
Thank you; it works :)

I have seen the build, but it doesn't running on my M1 Mac, so I tried to compile it by myself.

Issue #113, opened 2020-11-14
[Suggestion] Add buttons to simplify enhancements settings


Max833:
I think most people don't know that bsnes has a fast mode and an accurate mode and they are really confused when they see the enhancements settings. My suggestion is to add something that simplify this. byuu already did this with the driver settings:

![image](https://user-images.githubusercontent.com/53529402/99152995-cf0b0f80-26a5-11eb-94e6-9ccf04d3166b.png)

Maybe we can do the same for the enhancements window. Something like "Accurate mode", that disables all enhancements and "Fast mode" plus a short description.

Screwtapello:
This seems like a good idea! I would accept a patch for this.

PR #158, opened 2021-02-22
Add mode presets for accuracy and performance mode to simplify enhancements settings

bsnesFan:
https://user-images.githubusercontent.com/79463574/108725701-ed936900-7526-11eb-9216-dfb4935aad94.mp4

Fixes: https://github.com/bsnes-emu/bsnes/issues/113

When a game is loaded, you will get an information that you should reload the game in order for all changes to take effect. If no game is loaded, you will not see this, because the changes will take effect immediately. It's working really good and I think with that user will better unterstand what they have to enable for a good gaming experience.

Other than that, I'm not sure whats the best way to describe these two mode presets, but I don't think something like:
**Performance mode: Maximum speed, [...]**
is right, because it isn't _just_ about speed, that's why something like "Prefer HLE" is disabled. We should come up with something appropriate, so people want to use this mode by default. You can decide what's the best way to do this.

Thanks again for your help, Screwtape. :)

**Edit:** 
- added hotfixesSpacer
- added overclocking note
- minor adjustments

image

### DIFF
--- a/bsnes/target-bsnes/settings/enhancements.cpp
+++ b/bsnes/target-bsnes/settings/enhancements.cpp
@@ -133,6 +133,54 @@ auto EnhancementSettings::create() -> void {
   }).setChecked(settings.emulator.hack.hotfixes).onToggle([&] {
     settings.emulator.hack.hotfixes = hotfixes.checked();
   });
-
-  note.setText("Note: some settings do not take effect until after reloading games.");
+  
+  
+  ppuModeLabel.setText("Mode Presets:").setFont(Font().setBold());
+  ppuModeRequirements.setText(
+    "Accuracy Mode: Maximum hardware accuracy, disables performance shortcuts and all enhancements.\n"
+    "Performance Mode: Best compromise between hardware accuracy and performance."
+  );
+  accuracyMode.setText("Accuracy Mode").onActivate([&] {
+    runAhead0.setChecked(); settings.emulator.runAhead.frames = 0;
+    cpuClock.setPosition(0); cpuClock.doChange();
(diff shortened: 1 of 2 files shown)


Max833:
This is a very useful feature and it's working great. It's been a almost half a year.... do you think it is possible that someone can review that, @Screwtapello?

Kawa-oneechan:
I like it.

jchv (approved):
LGTM. Added information in the UI seems like a good idea, too.

bsnesFan:
Thank you :)

Shideravan (commented):
It's a nice improvement. For later improvement, maybe a third preset which also uses overclocking.

Max833:
> Other than that, I'm not sure whats the best way to describe these two mode presets, but I don't think something like:
> Performance mode: Maximum speed, [...]
> is right, because it isn't just about speed, that's why something like "Prefer HLE" is disabled. We should come up with something appropriate, so people want to use this mode by default. You can decide what's the best way to do this.

You are right. And I'm pretty sure performance mode will also have less visual bugs than accuracy mode overall (at least for now, see: https://arstechnica.com/gaming/2021/06/how-snes-emulators-got-a-few-pixels-from-complete-perfection/).

Besides, I've played the whole library with the scanline renderer and Near fixed _almost_ every bug (some games are not compatible and switch to the cycle-based renderer). A lot of shaders don't work with the cycle-based renderer as well. So... we should write a text that is appealing, so everyones first choice is the performance mode.

bsnesFan:
Thank you. I did overhaul my pull request.

PR #164, opened 2021-03-12
make it compilable on case-sensitive systems

kdridi:


jchv (approved):

Screwtapello:
Thanks for spotting that!

PR #174, opened 2021-04-04
Auto joypad cleanup

Screwtapello:
Jonas Quinn refactored higan's auto-joypad polling in https://github.com/higan-emu/higan/commit/e422ddc7293b10aae8a1935d4c7f96aeb8445430, then Asura re-implemented those changes for bsnes in 39c37ec. A bug was found in bsnes, fixed in 4f7a269, and Jonas Quinn ported the bsnes version of the code back to higan in https://github.com/higan-emu/higan/commit/9a625c545ca89b094d5c1da40bbfa5d07156a4aa.

In the process of that (back?) port, Jonas discovered that the `autoJoypadActive` and `autoJoypadLatch` variables were no longer actually being used in bsnes. Also, Jonas' original patch taught higan what happens when auto-joypad polling is disabled while active, and that detail had not made it across to bsnes before.

With these changes, bsnes and higan should once again be equally accurate emulators.

### DIFF
--- a/bsnes/sfc/cpu/cpu.hpp
+++ b/bsnes/sfc/cpu/cpu.hpp
@@ -118,8 +118,6 @@ private:
     bool hdmaPending = 0;
     bool hdmaMode = 0;  //0 = init, 1 = run
 
-    bool autoJoypadActive = 0;
-    bool autoJoypadLatch = 0;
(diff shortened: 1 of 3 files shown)


near-san:
Looks great, thank you for all your hard work Screwtape!! And Jonas too!

PR #176, opened 2021-04-11
hiro: Force the GTK+3 backend to always use X11.

Screwtapello:
In theory, the GTK+3 backend should automatically give us support for Wayland and other display protocols. In practice, hiro still has a lot of hard-coded calls to X11 APIs. Until we get those sorted out, let's make bsnes run in backward compatibility mode instead of crashing at startup.


Screwtapello:
This was already done in a different way.

PR #182, opened 2021-05-26
Fix vector<T>::acquire() arguments

Morilli:
This would fail to compile previously because `_pool`, which is of type `T*`, is assigned a variable of type `const T*` which results in a compiler error.
Note that this does not affect bsnes at all because this function is unused.


Screwtapello:
Thanks for reporting this!

PR #196, opened 2021-07-11
hiro: Defer menubar updates on Windows to reduce flicker.

Max833:
> Menu bar updates are deferred to the event loop. A special case is made when no menu bar is visible, to ensure one is visible as early as possible during application startup.

Ported from **jchv**:

https://github.com/higan-emu/higan/pull/146

Fixes bsnes menubar as well.


Max833:
I would like to do another pull request. Can someone review this one, @Screwtapello or @Kawa-oneechan? This would be really helpful, thank you. =)

PR #199, opened 2021-07-28
Fix build failure on UCRT mingw64

### LINKED ISSUE #198: <nall/windows/guard.hpp> is broken on ucrt
When compiling on ucrt64 mingw, I noticed I get some very weird errors about undefined symbols in stdlib.

I tracked it down to the `#define __MSVCRT_VERSION__ WINVER` line in `<nall/windows/guard.hpp>`, which presumably should "fix" some system headers, but because of this hardcoding instead breaks the ucrt build.

I'm not sure what the "right" fix here is, but I suggest
1. Removing this header altogether; it really doesn't feel right to me at all, or
2. at least remove all heuristic (re-)defines from it

Perhaps some insight on why this was added / needed? in the first place would be helpful.
Fwiw my current (system) define of `__MSVCRT_VERSION__` is `0xE00`.

Morilli:
closes #198.

Note that `#define __MSVCRT_VERSION__ WINVER` is the only thing that broke the build, but I've removed the others regardless because
1. it doesn't feel right to define some of these and not others
2. Looks like they're bound to cause issues in the future as well

Note that the resulting binary *does* change after removing `#define _WIN32_IE WINVER`... don't ask me why, I genuinely have no idea why the internet explorer is of any relevancy here. Chose to ignore and move on 👍

### DIFF
--- a/nall/windows/guard.hpp
+++ b/nall/windows/guard.hpp
@@ -5,12 +5,16 @@
 #define interface WindowsInterface
 
 #undef UNICODE
+#undef WINVER
 #undef WIN32_LEAN_AND_MEAN
+#undef _WIN32_WINNT
 #undef NOMINMAX
 #undef PATH_MAX
 
 #define UNICODE
+#define WINVER 0x0601
 #define WIN32_LEAN_AND_MEAN
+#define _WIN32_WINNT WINVER
 #define NOMINMAX
 #define PATH_MAX 260
 
--- a/nall/windows/guard.hpp
+++ b/nall/windows/guard.hpp
@@ -5,20 +5,12 @@
 #define interface WindowsInterface
 
 #undef UNICODE
-#undef WINVER
-#undef WIN32_LEAN_AND_LEAN
-#undef _WIN32_WINNT
-#undef _WIN32_IE
-#undef __MSVCRT_VERSION__
+#undef WIN32_LEAN_AND_MEAN
 #undef NOMINMAX
 #undef PATH_MAX
 
 #define UNICODE
-#define WINVER 0x0601
 #define WIN32_LEAN_AND_MEAN
-#define _WIN32_WINNT WINVER
-#define _WIN32_IE WINVER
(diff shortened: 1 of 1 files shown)


Screwtapello:
Hmm, apparently this breaks the Windows builds on both GitHub and CirrusCI. I guess *some* of those defines must be important.

Morilli:
From what I can tell so far, it is missing `WSAPoll`, which is defined in `winsock2.h` iff `_WIN32_WINNT >= 0x0600`.
On my system, there is a `_mingw.h` header that does `#define _WIN32_WINNT 0x601` and is included from `windows.h` which in turn gets included in `winsock2.h` iff `_INC_WINDOWS` is NOT defined.
`_INC_WINDOWS` only gets defined if `windows.h` was already included from what I can see.

That means the only reasonable explanation for the failure is that on whatever system the CI is running, `_WIN32_WINNT` gets defined to a value lower than `0x0600` in `_mingw.h` or even not at all.
According to [microsoft's documentation](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-160), the value `0x0600` corresponds to windows vista. Is the build running on an even older system (or in some compatibility mode or whatever)?

Morilli:
Okay, I have checked version `8.0.2` of `mingw-w64` (even though the CI uses 8.1, but I couldn't find that online), which defines `#define _WIN32_WINNT 0x502`.

I'm using msys2/mingw, which uses the current trunk version of mingw-w64 (which is version 9+) and defines what I wrote above.

I guess for the sake of supporting this *Windows Server 2003* define, I will re-add the WINNT and WINVER defines and see how it goes then.

Screwtapello:
I think it's a good idea to explicitly ask for the version of Windows we intend to support, regardless of what the toolchain defaults might happen to be.

Thanks for figuring this out!

Screwtapello:
I squashed these two commits and merged the result: 41f30f348a79387461a2091d1c1690a3c656396d

PR #200, opened 2021-07-31
Update to latest libco and SameBoy

carmiker:
This pull request updates bsnes to use the latest revision of libco and the latest tagged release of SameBoy.

Also included are the required build changes, version string update, and the addition of a log callback for SameBoy which suppresses logging to stdout when running GB content.


Screwtapello:
GitHub doesn't make it easy to explore a big PR like this, so I used the `update-subtrees.sh` script myself and merged the result: 3ccb54362220fb4aa4b7dced91eefcc733e351ce

Thanks for helping keep bsnes up to date!

PR #201, opened 2021-08-01
nall: File timestamp fix for OpenBSD and DragonFly BSD, add DragonFly BSD to PLATFORM_BSD definition

carmiker:
This is a manual backport of this commit from the upstream `nall` repository:
https://github.com/higan-emu/nall/commit/c673bd802e84050d83cb977955550947dccf9aef

Since the copy of `nall` in `bsnes` has diverged from the upstream repository, simply importing upstream `nall` is not possible without some changes deep in the `bsnes` codebase. Until those changes happen, this commit gives us out of the box OpenBSD support, while also paving the way for out of the box support for DragonFly.

### DIFF
--- a/nall/inode.hpp
+++ b/nall/inode.hpp
@@ -82,6 +82,12 @@ struct inode {
     #if defined(PLATFORM_WINDOWS)
     //on Windows, the last status change time (ctime) holds the file creation time instead
     case time::create: return data.st_ctime;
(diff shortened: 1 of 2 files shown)


Screwtapello:
Can you please file an issue about merging the upstream version of nall into bsnes, with whatever information you've gathered about the obstacles involved?

PR #219, opened 2021-09-06
Reorder optimal audio drivers

ghisvail:
Same order as defined in ares, which favor pulseaudio over ALSA.


ghisvail:
Keep in mind that `bsnes` will still load `safestDriver` over `optimalDriver` by default, so the current behavior on first launch will not change. This PR will help downstream distributions of `bsnes`, such as Linux distributions or [Flatpak](https://github.com/flathub/dev.bsnes.bsnes), start the app with more appropriate defaults by only patching `safestDriver` to `optimalDriver`.

PR #224, opened 2021-09-21
Allow Haiku build of libretro target

kwyxz:
For your consideration, this allows building the libretro target on the Haiku operating system. Haiku is mostly POSIX but doesn't have a full port of GTK hence standalone bsnes is a no-go, but the libretro target builds and works flawlessly.

### DIFF
--- a/nall/GNUmakefile
+++ b/nall/GNUmakefile
@@ -32,6 +32,8 @@ ifeq ($(platform),)
     platform := linux
   else ifneq ($(findstring BSD,$(uname)),)
     platform := bsd
+  else ifneq ($(findstring Haiku,$(uname)),)
+    platform := haiku
   else
     $(error unknown platform, please specify manually.)
   endif
@@ -137,6 +139,11 @@ ifeq ($(platform),bsd)
   options += -Wl,-rpath=/usr/local/lib
   options += -Wl,-rpath=/usr/local/lib/gcc8
   options += -lstdc++ -lm
+endif
+
+# haiku settings
+ifeq ($(platform),haiku)
+  options += -lroot
 endif
 
 # threading support
--- a/nall/directory.hpp
+++ b/nall/directory.hpp
@@ -267,19 +267,19 @@ inline auto directory::copy(const string& source, const string& target) -> bool 
   }
 #else
   inline auto directoryIsFolder(DIR* dp, struct dirent* ep) -> bool {
-#if defined(PLATFORM_HAIKU)
-    struct stat sp = {0};
-    stat(ep->d_name, &sp);
-    if S_ISDIR(sp.st_mode) return true;
-    if (S_ISLNK(sp.st_mode) || not S_ISREG(sp.st_mode)) {
-      fstatat(dirfd(dp), ep->d_name, &sp, 0);
-      return S_ISDIR(sp.st_mode);
-    }
-#else // not PLATFORM_HAIKU
+#if defined(PLATFORM_MACOS) || defined(PLATFORM_LINUX) || defined(PLATFORM_BSD)
     if(ep->d_type == DT_DIR) return true;
     if(ep->d_type == DT_LNK || ep->d_type == DT_UNKNOWN) {
       //symbolic links must be resolved to determine type
       struct stat sp = {0};
+      fstatat(dirfd(dp), ep->d_name, &sp, 0);
+      return S_ISDIR(sp.st_mode);
+    }
+#else // strictly POSIX systems
+    struct stat sp = {0};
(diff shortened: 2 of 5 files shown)


nall/GNUmakefile
```diff
@@ -139,6 +141,11 @@ ifeq ($(platform),bsd)
   options += -lstdc++ -lm
 endif
 
+# haiku settings
+ifeq ($(platform),haiku)
+  options += -lroot
```
Screwtapello, line 146: Out of curiosity, what does this do?

nall/directory.hpp
```diff
@@ -267,13 +267,23 @@ inline auto directory::copy(const string& source, const string& target) -> bool
...
   inline auto directoryIsFolder(DIR* dp, struct dirent* ep) -> bool {
+#if defined(PLATFORM_HAIKU)
+    struct stat sp = {0};
+    stat(ep->d_name, &sp);
+    if S_ISDIR(sp.st_mode) return true;
+    if (S_ISLNK(sp.st_mode) || not S_ISREG(sp.st_mode)) {
+      fstatat(dirfd(dp), ep->d_name, &sp, 0);
+      return S_ISDIR(sp.st_mode);
+    }
+#else // not PLATFORM_HAIKU
```
Screwtapello, line 278: It seems that POSIX only mandates the `d_name` and `d_ino` fields, not `d_type`. Rather than this being a special-case for `PLATFORM_HAIKU` it should probably be the default implementation for POSIXy OSs, and the existing implementation should be behind `#if defined(PLATFORM_MACOS) || defined(PLATFORM_LINUX) || defined(PLATFORM_BSD)` since those are the platforms we happen to know have this extra feature.

We should probably also have a fallback to the above code if `d_type == DT_UNKNOWN` (for when the filesystem doesn't support `d_type`) too, but you don't have to do that in this PR, I'm just thinking out loud.

ruby/GNUmakefile
```diff
@@ -35,6 +35,11 @@ ifeq ($(ruby),)
     $(eval $(call pkg_check,libpulse,audio.pulseaudiosimple))
     ruby += input.xlib
     $(eval $(call pkg_check,sdl2,input.sdl))
+  else ifeq ($(platform),haiku)
+    ruby += video.glx video.glx2
+    ruby += audio.oss
+    $(eval $(call pkg_check,openal,audio.openal))
+    $(eval $(call pkg_check,sdl2,input.sdl))
```
Screwtapello, line 42: If you can't build the bsnes desktop UI due to lack of GTK+ or equivalent, have you actually tested this code?

If Haiku doesn't have GTK+, does it have Qt?

nall/GNUmakefile
kwyxz: In the Haiku world, libroot is basically the equivalent of the glibc in Linux. You'll find more details here : https://www.haiku-os.org/docs/api/group__libroot.html

nall/directory.hpp
kwyxz: Indeed, d_type is the most common thing I need to look for when porting things to Haiku. I will turn the ifdef around as you proposed, it will also help for other future ports (another one of my pet peeves is porting libretro cores to Solaris, and this kind of things happens there as well).

ruby/GNUmakefile
kwyxz: I was quite positive this would not impact the libretro core but wanted to play it safe with libs I knew existed in Haiku.
Haiku has QT, though, I had not realized bsnes offered a QT interface. I can look into getting this to build.

kwyxz: Quick update : working on this, but since Haiku does not use X11 and has its own display things are a little bit hairier than I hoped for, as I am not really a developer :D but things are moving forward slowly.

kwyxz: Okay so here's the situation : building standalone bsnes on Haiku will require substantial code changes that I don't feel comfortable doing.
The QT flavor of hiro seems to be building fine, and is not the issue.
The issue is that Haiku does not have a proper shm.h implementation, which won't allow building the XShm or XVideo outputs even if I install X11 libs. 
It also does not implement the various glX functions bsnes uses (even though most are defined in the GLEW headers, they are not present in the Mesa libs in the system) which means no OpenGL output.
I suggest we stick with the libretro core right now. I will submit the other changes shortly.

Screwtapello:
Sorry for the long wait for approval! The changes you made look good, so I squashed them into a single commit and merged it: 6fc6bf14a39d32dab69c4f9687a81df26d412758

Thank you very much!

PR #241, opened 2022-03-14
fix Super FX voxel demo

jbo-85:
sfc: Add database entry for the "Super FX voxel demo" to specify a RA…M size of 64k instead of 32k
Support ROM/RAM addressing for Super FX ROMs that are not a power of 2
fixes 123

### DIFF
--- a/bsnes/Database/Super Famicom.bml
+++ b/bsnes/Database/Super Famicom.bml
@@ -30,6 +30,30 @@ game
       size: 0x80000
       content: Program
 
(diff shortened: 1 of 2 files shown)


carmiker:
Hey @jbo-85, I have experimented with these modifications after seeing you've committed a fix for a regression. It appears you don't actually need to do the RAM read/write changes.

jbo-85:
@carmiker 
Thanks, I reverted that

PR #274, opened 2023-03-07
nall: Fix the build for gcc13

orbea:
```
Compiling hiro/hiro.cpp ...
In file included from ../nall/arithmetic.hpp:37,
                 from ../nall/hash/hash.hpp:3,
                 from ../nall/hash/sha256.hpp:3,
                 from ../nall/file-buffer.hpp:12,
                 from ../nall/file.hpp:3,
                 from ../nall/directory.hpp:3,
                 from ../hiro/core/core.hpp:4,
                 from ../hiro/core/core.cpp:13,
                 from ../hiro/hiro.cpp:5:
../nall/arithmetic/natural.hpp: In function 'void nall::div(const uint256_t&, const uint256_t&, uint256_t&, uint256_t&)':
../nall/arithmetic/natural.hpp:239:23: error: 'runtime_error' is not a member of 'std'
  239 |   if(!rhs) throw std::runtime_error("division by zero");
      |                       ^~~~~~~~~~~~~
../nall/arithmetic/natural.hpp:1:1: note: 'std::runtime_error' is defined in header '<stdexcept>'; did you forget to '#include <stdexcept>'?
  +++ |+#include <stdexcept>
    1 | #define ConcatenateType(Size) uint##Size##_t
In file included from ../nall/arithmetic.hpp:45:
../nall/arithmetic/natural.hpp: In function 'void nall::div(const uint512_t&, const uint512_t&, uint512_t&, uint512_t&)':
../nall/arithmetic/natural.hpp:239:23: error: 'runtime_error' is not a member of 'std'
  239 |   if(!rhs) throw std::runtime_error("division by zero");
      |                       ^~~~~~~~~~~~~
../nall/arithmetic/natural.hpp:239:23: note: 'std::runtime_error' is defined in header '<stdexcept>'; did you forget to '#include <stdexcept>'?
In file included from ../nall/arithmetic.hpp:53:
../nall/arithmetic/natural.hpp: In function 'void nall::div(const uint1024_t&, const uint1024_t&, uint1024_t&, uint1024_t&)':
../nall/arithmetic/natural.hpp:239:23: error: 'runtime_error' is not a member of 'std'
  239 |   if(!rhs) throw std::runtime_error("division by zero");
      |                       ^~~~~~~~~~~~~
../nall/arithmetic/natural.hpp:239:23: note: 'std::runtime_error' is defined in header '<stdexcept>'; did you forget to '#include <stdexcept>'?
In file included from ../nall/arithmetic.hpp:61:
[...]

Issue #202, opened 2021-08-01
Errors importing upstream nall into bsnes repository


carmiker:
When attempting to import the upstream `nall` repository into `bsnes`, a large number of errors occur at compile time.

Although there is quite an onslaught of error text, it all boils down to just two errors repeating many times, all in `target-bsnes/program/platform.cpp`, `target-bsnes/program/game-pak.cpp`. and `target-bsnes/program/game-rom.cpp`:
```
target-bsnes/program/platform.cpp: In member function 'virtual nall::shared_pointer<nall::vfs::file> Program::open(uint, nall::string, nall::vfs::file::mode, bool)':
target-bsnes/program/platform.cpp:15:35: error: 'open' is not a member of 'nall::vfs::file'
   15 |       result = vfs::memory::file::open(Resource::System::Boards, sizeof(Resource::System::Boards));
```

```
In file included from target-bsnes/program/program.cpp:4:
target-bsnes/program/game-pak.cpp: In member function 'nall::shared_pointer<nall::vfs::file> Program::openPakSuperFamicom(nall::string, nall::vfs::file::mode)':
target-bsnes/program/game-pak.cpp:2:15: error: 'nall::vfs::fs' has not been declared
    2 |   return vfs::fs::file::open({superFamicom.location, name}, mode);
```

The solution to this is not obvious to me. Perhaps someone with a deeper understanding of `bsnes` and `nall` can offer some insight and lead the way to the most correct solution.

PR #286, opened 2023-06-08
nall: Add needed #include <stdexcept>

Screwtapello:
Taken from ares commit 6a7898396a14eef257e63cd002fb26ffbf6e2581

Apparently this is needed by GCC 13.

### DIFF
--- a/nall/arithmetic.hpp
+++ b/nall/arithmetic.hpp
@@ -3,6 +3,8 @@
 //multi-precision arithmetic
 //warning: each size is quadratically more expensive than the size before it!
 
+#include <stdexcept>
+
 #include <nall/stdint.hpp>
 #include <nall/string.hpp>
 #include <nall/range.hpp>


orbea:
I thought I already fixed this?

https://github.com/bsnes-emu/bsnes/pull/274

Screwtapello:
Ah, crud. Somebody reported this in higan, copying the fix from ares, and the same fix wasn't in bsnes so I assumed I'd have to port it there too. Sorry about that!

orbea:
The way files are included in nall it makes sense there are multiple viable places to put this include, ares has also continued to develop nall so it perhaps the fixes will not be 1:1 with their copy?

Screwtapello:
I think Near continued to develop nall in higan long before ares forked off, so trying to update nall to match the latest version would probably be quite the effort. I haven't actually tried it myself, though!

orbea:
There is an issue for that. https://github.com/bsnes-emu/bsnes/issues/202

PR #294, opened 2023-08-05
libretro: Support blargg's NTSC filter

### LINKED ISSUE #289: Feature Request: Please add Blarggs NTSC SNES filters
Is it possible to please add Blarggs NTSC filters options to BSNES like RF, Composite, S-Video and RGB?

These filters really make the games shine they way they were meant to be. 

Thank you for considering!

Edit*

@carmiker 
@Screwtapello 

I had a question for you guys as I know every once in a while you guys make commits with retroarch bsnes. 

Is bsnes constantly in some sort of weird 512x448 hi res mode or something even when most games should be in 256x224? 

The reason why I ask is because no NTSC filter shader seems to work on bsnes at all, even when you put it in rf which should really make everything "shimmery" if you know what I mean. 

It's almost like it ignores any NTSC aspect from shaders that have it. 

Being able to combine blarggs rf or composite filter would be great.

Hopefully I'm making sense.

carmiker:
This adds support for all four of blargg's NTSC filter options to the libretro port of bsnes.

As a consequence of using the filter functionality, the pixel format has been changed to XRGB8888 (preferable to the awkward 0RGB1555 when considering cross-platform compatibility and it is seemingly deprecated anyway). It also now builds on musl-libc without patches.

Closes #289

### DIFF
--- a/bsnes/target-libretro/libretro.cpp
+++ b/bsnes/target-libretro/libretro.cpp
@@ -287,7 +287,32 @@ static void flush_variables()
 		else
 			run_ahead_frames = atoi(variable.value);
 	}
-	
+
+	variable = { "bsnes_video_filter", nullptr };
+	if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &variable) && variable.value)
+	{
+		if (strcmp(variable.value, "None") == 0) {
+			program->filterRender = &Filter::None::render;
+			program->filterSize = &Filter::None::size;
+		}
+		else if (strcmp(variable.value, "NTSC (RF)") == 0) {
+			program->filterRender = &Filter::NTSC_RF::render;
+			program->filterSize = &Filter::NTSC_RF::size;
+		}
+		else if (strcmp(variable.value, "NTSC (Composite)") == 0) {
+			program->filterRender = &Filter::NTSC_Composite::render;
+			program->filterSize = &Filter::NTSC_Composite::size;
+		}
+		else if (strcmp(variable.value, "NTSC (S-Video)") == 0) {
+			program->filterRender = &Filter::NTSC_SVideo::render;
+			program->filterSize = &Filter::NTSC_SVideo::size;
+		}
+		else if (strcmp(variable.value, "NTSC (RGB)") == 0) {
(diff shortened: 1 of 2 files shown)


joepogo:
Wow!! Thank you so much @carmiker !!! 😁 I guess it'll take a little bit to get this committed and then it'll need to be merged with libretro right? I'm wondering I'm the meantime if it's possible to try to compile the dll myself to test this out on retroarch. Thank you so much man! It's great to be able to have these built in and make the classics really shine!!

carmiker:
If you post the errors, perhaps the problem can be solved. There is nothing in this commit that should break the build.

greenchili2:
Retroarch libretro BSNES Git is here.. https://github.com/libretro/bsnes-libretro/tree/master/bsnes and although bsnes-emu/bsnes is the master the Retroarch Libretro git is 99 commits ahead and 1 commit behind here.

https://github.com/libretro/bsnes-libretro/blob/master/bsnes/target-libretro/libretro.cpp
https://github.com/libretro/bsnes-libretro/blob/master/bsnes/target-libretro/program.cpp

There are numerous changes as well as changes in the names of the retropad defines, etc.  ( and who knows what else ).

Screwtapello:
I can't easily test this myself, since I don't use RetroArch. I was a little worried that the code seemed to be adding calls to blargg's filtering code but not actually adding the filtering code to the build system... but apparently it has *always* linked the filtering code even though nothing has ever used it before? Weird, but as long as it works, I guess.

greenchili2:
> I can't easily test this myself, since I don't use RetroArch. I was a little worried that the code seemed to be adding calls to blargg's filtering code but not actually adding the filtering code to the build system... but apparently it has _always_ linked the filtering code even though nothing has ever used it before? Weird, but as long as it works, I guess.

That is what it looked like to me as well.  That it was linked but some of the code that selects the filter used was removed and it defaulted to none?

carmiker:
The filter was never actually used at all, it was just using the raw pixel buffer directly out of the PPU. The "None" filter type merely converts the pixel buffer to a 32-bit format from 16-bit. Why was it being built? I'm not really sure either, but it appears the libretro build is piggybacking on other parts of the build system.

PR #300, opened 2023-10-30
wdc65816: fix PLB wrapping in emulation mode

Morilli:
Cherry-picked https://github.com/ares-emulator/ares/commit/e589f8d9cf1aaf04b1959bfc7d7085cb3861149c.
I've checked and confirmed this does fix the test in question.
Note that this does not fix the additional tests from [version 1.2](https://github.com/gilyon/snes-tests/releases/tag/v1.2) of the cpu test rom.

### DIFF
--- a/bsnes/processor/wdc65816/instructions-other.cpp
+++ b/bsnes/processor/wdc65816/instructions-other.cpp
@@ -205,9 +205,10 @@ E S.h = 0x01;
 auto WDC65816::instructionPullB() -> void {
   idle();
   idle();
-L B = pull();
+L B = pullN();
   ZF = B == 0;
   NF = B & 0x80;
+E S.h = 0x01;
 }
 
 auto WDC65816::instructionPullP() -> void {


Screwtapello:
For my own future reference, the current version of bsnes passes all the tests in `spctests.sfc` v1.2. For `cputests.sfc` v1.2, it fails tests 0027, 0074, 015F, 0194, 0216, 02CC, 030B, 0397, 03D9, 046F, 04E0, 0518, 05D2, 0621 — all the tests listed as testing undocumented behaviour.

On this branch, test 03d9 now passes, which is the PLB test, which is good.

I also saw test 0194 intermittently passing, but I can't figure out how to reproduce that glitch, so I'm going to ignore it for now. Thanks for contributing!

orbea:
@Screwtapello Just for reference the 0194 test doesn't fail every run, this indicates that there is something wrong in the test or that bsnes has indeterminate behavior.

PR #306, opened 2023-11-15
CI: Remove uses of the upload-release-asset action.

Screwtapello:
This action is deprecated, and was producing warnings every time the actions were run.


Screwtapello:
Closed in favour of #307.

PR #307, opened 2023-11-16
Simplify + update GitHub Actions workflow

jchv:
- Update steps to latest action versions
- Use GitHub script (Node.JS + Octokit.js) step to:
    - Try to delete old release.
    - Try to delete old tag.
    - Create new nightly tag for release.
- Use `softprops/action-gh-release` to actually create the releases.
    - GitHub has officially recommended this action in the deprecated official `upload-release-asset` action.
    - It seems like what it does is complicated enough to warrant actually using a third-party action instead of rolling our own.

The hope is that using Octokit.js will buy us a lot of time, as the API should be very stable at this point in time. Even if the GitHub API shifts, these fairly high-level calls shouldn't need to break much if at all, and I think the stuff this script does are complicated enough to warrant avoiding bash for the main logic. Using Octokit.js theoretically gives us retry logic and more for free, which should make the code less prone to flaking out.

This is mostly just a proposal, and I'd be happy to make changes to try to make it more palatable. We could try to move the release creation into the script step if we want to avoid needing a third-party action altogether, though I am skeptical it would be worth it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
     tags: [ 'v*' ]
   pull_request:
     branches: [ master ]
+permissions:
+  contents: write
 jobs:
   build:
     strategy:
@@ -18,7 +20,7 @@ jobs:
           version: latest
     runs-on: ${{ matrix.os.name }}-${{ matrix.os.version }}
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Install Dependencies
       if: matrix.os.name == 'ubuntu'
       run: |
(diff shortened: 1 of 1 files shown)


Screwtapello:
I do wonder whether it would be better to use the third-party `softprops/action-gh-release@v1` action, or just have another instance of the `actions/github-script@v7` action and write the logic in JS ourselves... but that's really a question about which of the two is more likely to break first, and I do not have a sufficiently-informed opinion to guess, let alone decide.

I'll leave this open for a day or two to see if anybody else has feedback, but if nobody else objects I'll merge it (unless somebody else beats me to it).

jchv:
We could probably just do it all in a single github-script action. I can look into it.

PR #331, opened 2024-09-21
Implement SA1 I-RAM write protection

Morilli:
Logic is based on [what ares does](https://github.com/ares-emulator/ares/blob/d6858ad8219a031a02dd4c91f90f69be6349b1ee/ares/sfc/coprocessor/sa1/iram.cpp).

I've checked with the testrom from https://github.com/absindx/SNES-TestRoms/tree/a229ee7215effdd1d21ef70dbcb15694ac4d8046/SA1RamProtectionTest and have confirmed that bsnes now passes all tests.


orbea:
This was tested and merged into bsnes-jg, thanks!

https://gitlab.com/jgemu/bsnes/-/commit/d4129eccf56e15642215159722fb11de21483eed

PR #332, opened 2024-09-24
Database: Add entries for Competition carts

carmiker:
Entries are added to Super Famicom.bml for the Campus Challenge '92 and PowerFest '94 competition cartridges. Additionally, the mask for the NEC DSP on the EVENT-CC92 board is modified as per Snes Central: https://snescentral.com/article.php?id=0790


carmiker:
Special thanks to the ares contributors who did some of the legwork here: https://github.com/ares-emulator/ares/pull/1648

This PR in its current state will still need some work to be fully useful. There will need to be some file loading routines to handle splitting up concatenated ROM images and placing them into the proper memory buffers inside the emulator. The ability to change DIP switch settings will also allow setting the default 6 minute game time used in the real competitions (default is 3 if the DIP switches are set to 0).

Screwtapello:
As I understand it (from my discussion with carmiker on IRC), bsnes can already play CC92 and PF94 if provided with a "game folder" with a pre-made `manifest.bml` that identifies the individual ROM files and the board to plug them into.

This PR (in its initial state) adds pre-made manifests to the game database with the SHA1 of the concatenated files, so if you have a copy in concatenated form ("Combined" in No-Intro terminology, instead of "Split"), bsnes knows what manifest to use with it.

The missing part is code to split the concatenated ROM into its individual parts during loading, the way that bsnes will split a combined "Super Mario Kart + DSP1b" ROM. Also (as noted), it would be nice to provide a way to change the DIP switches, or at least make the default 6 minutes instead of 3.

joepogo:
Thanks for looking at this guys! I can get powerfest 94 to work on the older bsnes accuracy core through bml in retroarch but neither powerfest 94 nor campus challenge will work on current bsnes using the rom folder with bml loading or the concatenated rom. 

I think near removed support for it for some years ago. I can only get powerfest 94 to work on retroarch using an old accuracy core. 

It'll be awesome to see this get implemented and hopefully afterwards get the retroarch current bsnes merged with this upstream.

Moizes82:
How do I use hotkeys for 200% and 100% speed?

joepogo:
Hey guys, just checking in. Is there still a possibilty of this getting fixed for current bsnes?

joepogo:
Thanks for putting this info here @carmiker ! I was able to get these going in bsnes libretro on my personal dll.core and everything looks good!  I hardcoded my time to 6 minutes in event.cpp for the time being as it was defaulting to 3 minutes.

PR #351, opened 2025-06-22
Improve auto-joypad polling timing

### LINKED ISSUE #322: "Wolverine - Adamantium Rage (USA)" hangs on the latest master version
As reported by user usVSthem on Discord, the USA version of the game *Wolverine - Adamantium Rage* hangs after displaying the unskippable developer and publisher logos, but before displaying the title screen. The European version, and the Japanese version (just named "Wolverine") both work fine.

Bisecting history points to 4f7a269ba5d053bb824b5ffa0978799331a5f2d8 as the commit that introduced the problem, which changed how bsnes handled joypad polling - which makes sense, since it's only once the game tries to interact with the joypad that the problem occurs.

Note that this commit fixes the game *SpellCraft - Aspects of Valor*, so whatever we do here has to make both games work - we can't just revert this specific fix.

Morilli:
This fixes #322.

In addition, these changes improve results in a number of test roms from https://github.com/bsnes-emu/bsnes/issues/322#issuecomment-2952297191.

Most of these test roms aren't "good or bad" and instead show trends and general behavior. In some cases it's clear when the emulator behavior is wrong and in other cases it's not so clear. The displayed values are not fixed and change frame-by-frame. I think it is most important to *roughly* match the hardware result. In some cases there may not even be one true correct result as even different hardware differs in results.

That being said, on current master the following tests are **clearly bad**:
```
01 auto-joy-timing-test
05 blip-autojoy-test-automatic
06 enable-autojoy-late-test-2
07 blip-autojoy-test
08 joypad-latch-set-during-autojoy
09 joypad-latch-clear-during-autojoy
10 joyser0-read-during-autojoy
12 blip-autojoy-timing-test
13 blip-autojoy-latches-joypad-test
14 clear-autojoy-after-autojoy-active
```

With the changes in this PR, all of these tests seem "fixed", at least mostly.

Notable remaining differences with hardware:
- Test 04 now has more `FFFF` in the column on the right. Should be 3-6 from the top, is 4-8.
- The right-most column in test 06 is a solid block of `FFFF` and does not flicker with occasional `0000` (it does on hardware).
- When holding `B` in test 08, the fourth row under `JOY 1` will never show `FFFF`, it shows `BFFF` at most.
- Similar to above, in test 09 when holding `B` the third row under `JOY 1` can never show `C000`, only `E000` and `F000`.
- When holding a key like `Y` in test 10, the left-most column under `JOY 1` may show 0s, but the video from hardware only shows 4s and 8s.

These are mainly minor off-by-1 problems, and roughly half of these also show in Mesen. I think the current results are good enough for now.

### DIFF
--- a/bsnes/sfc/cpu/serialization.cpp
+++ b/bsnes/sfc/cpu/serialization.cpp
@@ -45,6 +45,9 @@ auto CPU::serialize(serializer& s) -> void {
 
   s.integer(status.autoJoypadCounter);
 
+  s.integer(status.autoJoypadPort0);
+  s.integer(status.autoJoypadPort1);
+
   s.integer(io.wramAddress);
 
   s.boolean(io.hirqEnable);
--- a/bsnes/sfc/cpu/serialization.cpp
+++ b/bsnes/sfc/cpu/serialization.cpp
@@ -45,8 +45,8 @@ auto CPU::serialize(serializer& s) -> void {
 
   s.integer(status.autoJoypadCounter);
 
-  s.integer(status.autoJoypadPort0);
   s.integer(status.autoJoypadPort1);
+  s.integer(status.autoJoypadPort2);
 
   s.boolean(status.cpuLatch);
   s.boolean(status.autoJoypadLatch);
--- a/bsnes/sfc/cpu/serialization.cpp
+++ b/bsnes/sfc/cpu/serialization.cpp
@@ -48,6 +48,9 @@ auto CPU::serialize(serializer& s) -> void {
   s.integer(status.autoJoypadPort0);
   s.integer(status.autoJoypadPort1);
 
+  s.boolean(status.cpuLatch);
+  s.boolean(status.autoJoypadLatch);
+
   s.integer(io.wramAddress);
 
   s.boolean(io.hirqEnable);
--- a/bsnes/sfc/cpu/timing.cpp
+++ b/bsnes/sfc/cpu/timing.cpp
@@ -232,20 +232,23 @@ auto CPU::joypadEdge() -> void {
     }
   }
 
-  if(status.autoJoypadCounter >= 2 && !io.autoJoypadPoll) {
-    // if auto-joypad polling is disabled at this point skip the rest of the polling
-    status.autoJoypadCounter = 33;
-    return;
-  }
+  if(status.autoJoypadCounter >= 2) {
+    if (!io.autoJoypadPoll) {
+      // if auto-joypad polling is disabled at this point skip the rest of the polling
+      status.autoJoypadCounter = 33;
+      return;
+    }
 
-  if(status.autoJoypadCounter >= 2 && (status.autoJoypadCounter & 1)) {
     //sixteen bits are shifted into joy{1-4}, one bit per 256 clocks
-    uint2 port0 = controllerPort1.device->data();
-    uint2 port1 = controllerPort2.device->data();
-
-    io.joy1 = io.joy1 << 1 | port0.bit(0);
-    io.joy2 = io.joy2 << 1 | port1.bit(0);
-    io.joy3 = io.joy3 << 1 | port0.bit(1);
-    io.joy4 = io.joy4 << 1 | port1.bit(1);
+    //the bits are read on one 128-clock cycle and written on the next
+    if ((status.autoJoypadCounter & 1) == 0) {
+      status.autoJoypadPort0 = controllerPort1.device->data();
+      status.autoJoypadPort1 = controllerPort2.device->data();
+    } else {
+      io.joy1 = io.joy1 << 1 | status.autoJoypadPort0.bit(0);
+      io.joy2 = io.joy2 << 1 | status.autoJoypadPort1.bit(0);
+      io.joy3 = io.joy3 << 1 | status.autoJoypadPort0.bit(1);
+      io.joy4 = io.joy4 << 1 | status.autoJoypadPort1.bit(1);
+    }
   }
 }
--- a/bsnes/sfc/cpu/timing.cpp
+++ b/bsnes/sfc/cpu/timing.cpp
@@ -202,21 +202,20 @@ auto CPU::dmaEdge() -> void {
 
 //called every 128 clocks from inside the CPU::stepOnce() function
 auto CPU::joypadEdge() -> void {
-  //it is not yet confirmed if polling can be stopped early and/or (re)started later
-  if(!io.autoJoypadPoll) return;
-
-  if(vcounter() == ppu.vdisp() && hcounter() >= 130 && hcounter() <= 256) {
+  if(vcounter() == ppu.vdisp() && (counter.cpu & 255) == 0 && hcounter() >= 130 && hcounter() <= 384) {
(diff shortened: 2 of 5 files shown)


bsnes/sfc/cpu/serialization.cpp
```diff
@@ -45,6 +45,12 @@ auto CPU::serialize(serializer& s) -> void {
 
   s.integer(status.autoJoypadCounter);
 
+  s.integer(status.autoJoypadPort0);
+  s.integer(status.autoJoypadPort1);
+
+  s.boolean(status.cpuLatch);
+  s.boolean(status.autoJoypadLatch);
+
```
Screwtapello, line 53: If we change what's being serialised, we should also bump the serialisation version.

Morilli: You're right, and that's one thing I was contemplating while making the serialization and struct changes. But I looked at the last time serialization was changed (34af95da9ea8ec2b5957f62ff3265c09628192fc) and the version wasn't bumped after that either. Should I just bump it to something like `115.1`?

bsnes/sfc/cpu/timing.cpp
```diff
@@ -202,45 +202,54 @@ auto CPU::dmaEdge() -> void {
...
+    controllerPort1.device->latch(status.autoJoypadLatch | status.cpuLatch);
+    controllerPort2.device->latch(status.autoJoypadLatch | status.cpuLatch);
+  }
 
-    //shift registers are cleared to zero at start of auto-joypad polling
-    io.joy1 = 0;
-    io.joy2 = 0;
-    io.joy3 = 0;
-    io.joy4 = 0;
+  if(status.autoJoypadCounter != 1 && !io.autoJoypadPoll) {
```
Screwtapello, line 236: Why is the `status.autoJoypadCounter != 1` condition here? If the counter is 0 or 2-33, and auto-polling is disabled, we can skip the rest of the process, but if the counter is 1, we just wait until the counter becomes 2 and *then* disable auto-polling? If a game enables auto-polling after the counter ticks 1, and we've skipped the latching inputs and clearing the shift registers, does the SNES start shifting a new copy of the previously-latched inputs into the shift registers? Does this not happen if the game enables auto-polling on ticks 2 or later?

Morilli: You're right that this seems peculiar but it does appear to be accurate. Without this condition the tests `12-blip-autojoy-timing-test` and `13-blip-autojoy-latches-joypad-test` no longer match hardware results.

The implication of this condition is that disabling auto-joypad polling during this one cycle has no effect, so it could for example be toggled off and toggled on again without affecting the auto-joypad polling sequence at all during this specific timing window.

As far as I understand that's precisely what the blip-autojoy-timing-test intends to test: It waits until `HVBJOY` reports auto-joypad polling to be active, disables auto-joypad polling per `NMITIMEN` write, waits a certain amount of cpu cycles and then re-enables it. Of course this depends on `HVBJOY` and `NMITIMEN` behavior, but changing those as well causes other tests to break which leads me to believe that this combination is most accurate.

undisbeliever: Fiskbit and I noticed that temporary disabling (`NMITIMEN = 0`) then quickly re-enabling auto-joy (`NMITIMEN = 1`) would not stop automatic joypad reading.

`12-blip-autojoy-timing-test` was used to determine if is a 128-clock or 256-clock window from when `HVBJOY` reports auto-joypad active.  My results indicate it is a 256-clock window.  If the enable auto-joy `NMITIMEN` write is > 256-clocks after auto-joypad starts then automatic joypad read stops.

Quoting my notes from that day:
> It's matches what I expected for the 3chip.  I didn't know if it was a 128 clock or 256 clock before autojoy could be re-enabled.
> If my math is right, there's N + 6 fastROM cycles (48 mcycles) between HVBJOY reports auto-joy active and the NMITIMEN=1 write.
>
> * 206 delay = 254 m-cycles
> * 208 delay = 256 m-cycles.

<br/>

For the `13-blip-autojoy-latches-joypad-test`, Fiskbit asked for a test that checks if the S-CPU latches the controller twice if the `NMITIMEN` auto-joypad enable flag is blipped (disabled then quickly enabled).  Details about the test can be found in [the test's source code](https://github.com/undisbeliever/snes-test-roms/blob/master/src/hardware-tests/auto-joypad/blip-autojoy-latches-joypad-test.asm).

Morilli:
> Is that more or less correct?

Yes, that seems like a good description of the changes. I realize I could have explained more in the PR description, but I'm glad it was understandable regardless. Just a couple notes:

> what seems to happen in hardware is that we read one bit, then 128 clocks later we shift it in, then 128 clocks later we read the next bit. [...] I'm a little surprised that any software would notice the difference

It's possible this is coincidental due to the fact that the game spams `HVBJOY` polls and auto-joypad poll result reads. Perhaps someone engineered the mechanism in this way so it would always work on hardware or it was sheer luck.
The way the code seems to work is that `HVBJOY` is polled, and if it says auto-joypad polling isn't active, `0x4218`, `0x4219` and `0x421A` reads will be performed.

However there is a delay between checking the auto-joypad polling state and the actual read (TOCTOU problem). In rare cases, the last `HVBJOY` read will return auto-joypad polling to be inactive, but the next `io.joy` read isn't until after 3 auto-joypad polling cycles later, at which point `autoJoypadCounter` is 2. If `io.joy` was already written to in this cycle the game would read incorrect data and misbehave.

Here's an excerpt of a printf-log I created showcasing the problematic sequence of events:
```
[...] (hundreds of HVBJOY polls and io.joy reads)
HVBJOY | counter: 33 (reported inactive)
reading io.joy2
autoJoypad: 0
latched active and cleared io.joy
autoJoypad: 1
latched inactive
autoJoypad: 2
read data
reading io.joy1
autoJoypad: 3
wrote data
[...]
HVBJOY | counter: 9 (reported active)
```


> If auto-polling is disabled by the time we get past the latch interactions, we skip the rest of the auto-polling and re-enabling it won't do anything until the next frame.

This is true, but this is also true for the very start of auto-joypad polling (status.autoJoypadCounter == 0). If the sequence was just started and io.autoJoypadPoll is false then the controller latch will be closed (if it isn't already) and `status.autoJoypadCounter` will be set to 33 immediately.

Screwtapello:
Thank you to undisbeliever for doing the research, and to Morilli for figuring out how to update bsnes to match!

Morilli:
> Looks like Zenkoku Koukou Soccer 2 still has issues (https://github.com/ares-emulator/ares/issues/970)

That's the corresponding issue in ares, not bsnes. This problem is fixed in bsnes but someone would have to port the changes to ares for it to be fixed there as well.

Screwtapello:
OK, I tested this game in the latest bsnes nightly myself, and once I figured out how to get in-game, I repeatedly held the Start button for about a second, and recorded how many times the game stayed paused ("solid"), and how many times it immediately unpaused ("flaky"):

- total presses: 50
- solid pauses: 36
- flaky pauses: 14

My friend with the real SNES reports that it behaved very similarly.

PR #359, opened 2025-07-27
Update SameBoy to 1.0.2

alice-mkh:
SameBoy has unreachable() with parentheses now, as opposed to unreachable without them in nall. To work around the conflict, include SameBoy stuff as the first thing in its respective header, and undef unreachable afterwards.


alice-mkh:
One thing I'm not sure about:

> All files and directories in this repository, except for the iOS and HexFiend
> directories, are licensed under the Expat License:

@LIJI32 probably easiest to ask you - do these directories need to be excluded? I'm not sure because bsnes doesn't use them, but they are still there

alice-mkh:
Alright, thanks!

alice-mkh:
I didn't notice 1.0.2 got re-tagged to include https://github.com/LIJI32/SameBoy/commit/5b883465370bc7ad5a3576d26e7cee2ec553d3a0, updating again

LIJI32:
It's not a huge deal since bsnes doesn't use `cheats.c`. It might be a good idea to build SameBoy with these defines:
* `GB_DISABLE_REWIND`
* `GB_DISABLE_DEBUGGER`
* `GB_DISABLE_CHEATS`
* `GB_DISABLE_CHEAT_SEARCH`

And omit the following files:
* `rewind.c`
* `debugger.c`
* `sm83_disassembler.c`
* `symbol_hash.c`
* `cheats.c`
* `cheat_search.c`

These will allow better compiler optimizations, and also make the binary smaller which allows more efficient CPU cache usage. (Ideally, bsnes can also switch to using SameBoy as a library by building `libsameboy.a` with the relevant flags and using it as is. SameBoy's makefile will use the optimal compiler flags for the core this way)

PR #361, opened 2025-08-30
Remove Qt 4 support.

jchv:
Qt 4.8 LTS was released on December 15th, 2011. It hit its official end-of-life on December 31st, 2015. It has been dead for nearly 10 years. Qt 4 is now so old, that it has been removed from Debian long enough that "qt4" doesn't show up in *any* of the indexed package suites in Debian's online package search service.

With that in mind, I think we can safely say nobody is using this, nobody will miss this, and it is beyond time to get rid of it (and probably add Qt 6 support some day.)

Mirrored in:
- ares: https://github.com/ares-emulator/ares/pull/2187
- higan: https://github.com/higan-emu/higan/pull/216


Screwtapello:
Thanks for looking into this!

PR #366, opened 2025-09-01
hiro/qt: Remove deprecated usage of QApplication::hasPendingEvents

jchv:
Qt 6 no longer allows us to poll for the status of the event loop. However, Qt 6 does still have a `processEvents` function, and the overload with the timeout actually spins the event loop and returns as early as there are no more events in it. That's basically what we want to do. We could run this with a really huge timeout, but that would just make it more likely that hiro freezes up, so instead let's mirror hiro/gtk3 and use a 50ms timeout.

This also removes the public `Application::pendingEvents` API which does not seem like it was used by bsnes/higan/ares anyways.

### DIFF
--- a/hiro/core/application.cpp
+++ b/hiro/core/application.cpp
@@ -38,10 +38,6 @@ auto Application::run() -> void {
   return pApplication::run();
 }
 
-auto Application::pendingEvents() -> bool {
-  return pApplication::pendingEvents();
-}
-
 auto Application::processEvents() -> void {
   return pApplication::processEvents();
 }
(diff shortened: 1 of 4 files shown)


Alcaro:
- Someone should inline pApplication::pendingEvents() in the other backends into their callers. (But not necessarily you, if you can't test them.)
- Comparing the code's behavior to prior versions usually goes in commit message and/or PR description, not comments. (Not sure what the rules are in this specific project, though.)

Screwtapello:
I modified the comment to remove references to historical behaviour. I removed `pendingEvents()` from the GTK+ backend in b7678f52611325b6396a72498fde81da3ae58d0d and tested it, and filed #370 for removing it from the other backends (so we don't forget).

jchv:
>     * Comparing the code's behavior to prior versions usually goes in commit message and/or PR description, not comments. (Not sure what the rules are in this specific project, though.)

I suppose that's not an unreasonable point but if it's really that important I think it can be addressed by just dropping the last line of the comment some time.

PR #372, opened 2025-09-02
Remove GTK+2 support.

Screwtapello:
Although it had HIRO_GTK==3 sections, hiro/gtk/monitor.cpp was actually using deprecated GTK+2 APIs on GTK3, and the GTK3 code was broken and wouldn't compile.

I only have a single, standard-resolution monitor here to test with, I would appreciate testing from people who have multiple monitors, and hi-DPI monitors too since I modified that code.


jchv (approved):
Seems to work at least as well as it did before.

PR #377, opened 2026-02-11
CX4 program cache behaviour improvements

AkiteruSDA:
- When triggering a cache preload by writing to $7f48, update the pb register with the prepared program page
- When preloading a page, ignore all runtime checks and reload the page immediately

I have been looking into CX4 emulation lately, as it historically hasn't gotten a ton of attention compared to other chips and has been known to run quite fast compared to original hardware. By observing how Mega Man X2/X3 sets up the program cache on the CX4, I believe these two changes are consistent with assumptions made by the games on the CX4's caching behaviour.

I have created a [comparison video](https://www.youtube.com/watch?v=Laj1P41fXzU) of the Mega Man X2 opening sequence, showing the difference between current `master`, my commit `6c5f03e`, and footage recorded from my SNES and original cart. These changes result in a significant accuracy improvement on the wireframe head section of the opening.

I am certainly open to feedback on these changes as I am not a hardware expert. Just looking to do what I can to improve emulation accuracy for X2/X3. Thanks!

### DIFF
--- a/bsnes/processor/hg51b/hg51b.cpp
+++ b/bsnes/processor/hg51b/hg51b.cpp
@@ -89,6 +89,7 @@ auto HG51B::cache() -> bool {
   for(uint offset : range(256)) {
     step(wait(address));
     programRAM[io.cache.page][offset]  = read(address++) << 0;
+    step(wait(address));
     programRAM[io.cache.page][offset] |= read(address++) << 8;
   }
   return io.cache.enable = 0, true;
--- a/bsnes/processor/hg51b/hg51b.cpp
+++ b/bsnes/processor/hg51b/hg51b.cpp
@@ -72,15 +72,18 @@ auto HG51B::suspend() -> void {
 auto HG51B::cache() -> bool {
   uint24 address = io.cache.base + r.pb * 512;
 
-  //try to use the current page ...
-  if(io.cache.address[io.cache.page] == address) return io.cache.enable = 0, true;
-  //if it's not valid, try to use the other page ...
-  io.cache.page ^= 1;
-  if(io.cache.address[io.cache.page] == address) return io.cache.enable = 0, true;
-  //if it's not valid, try to load into the other page ...
-  if(io.cache.lock[io.cache.page]) io.cache.page ^= 1;
-  //if it's locked, try to load into the first page ...
-  if(io.cache.lock[io.cache.page]) return io.cache.enable = 0, false;
+  if(io.cache.preload == 0) {
+    //try to use the current page ...
+    if(io.cache.address[io.cache.page] == address) return io.cache.enable = 0, true;
+    //if it's not valid, try to use the other page ...
+    io.cache.page ^= 1;
+    if(io.cache.address[io.cache.page] == address) return io.cache.enable = 0, true;
+    //if it's not valid, try to load into the other page ...
+    if(io.cache.lock[io.cache.page]) io.cache.page ^= 1;
+    //if it's locked, try to load into the first page ...
+    if(io.cache.lock[io.cache.page]) return io.cache.enable = 0, false;
+  }
+  io.cache.preload = 0;
 
   io.cache.address[io.cache.page] = address;
   for(uint offset : range(256)) {
--- a/bsnes/processor/hg51b/hg51b.hpp
+++ b/bsnes/processor/hg51b/hg51b.hpp
@@ -152,6 +152,7 @@ protected:
     struct Cache {
       uint1  enable;
       uint1  page;
+      uint1  preload;
       uint1  lock[2];
       uint24 address[2];  //cache address is in bytes; so 24-bit
       uint24 base;        //base address is also in bytes
--- a/bsnes/processor/hg51b/registers.cpp
+++ b/bsnes/processor/hg51b/registers.cpp
@@ -12,13 +12,13 @@ auto HG51B::readRegister(uint7 address) -> uint24 {
   case 0x2e:
     io.bus.enable  = 1;
     io.bus.reading = 1;
-    io.bus.pending = 1 + io.wait.rom;
+    io.bus.pending = io.wait.rom;
     io.bus.address = r.mar;
     return 0x000000;
   case 0x2f:
     io.bus.enable  = 1;
     io.bus.reading = 1;
-    io.bus.pending = 1 + io.wait.ram;
+    io.bus.pending = io.wait.ram;
     io.bus.address = r.mar;
     return 0x000000;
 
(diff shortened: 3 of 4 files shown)


Screwtapello:
This looks plausible, and your video comparison is convincing. There's a lot of experienced SNES researchers watching this repo, though, I'll give them a day or two to think about it and chime in if they have objections or improvements.

AkiteruSDA:
Thank you for the response! I tested this again while turning **off** fast mode for coprocessors, and that caused this to start running fast. That seems like a strange result, so I want to make sure I fully understand why that might happen. Maybe it's really sensitive to how the chip is clocked? I'll try to look into it more.

carmiker:
@AkiteruSDA, do you mean with your change it speeds up with fast mode for coprocessors turned off? Was it the correct speed before the change with fast coprocessor mode turned off?

AkiteruSDA:
It speeds up with fast coprocessor mode turned off both before and after my change. Before my change, both fast coprocessor mode turned on and off were running fast. Here are some frame counts for each combination, timed from the first black frame after the Capcom logo to the first black frame after the Rockman X2 logo fades out:
```
master, fast coprocessor off: 4380
master, fast coprocessor on:  4397
6c5f03, fast coprocessor off: 4396
6c5f03, fast coprocessor on:  4412 (correct speed)
```

Edit: The CX4 setup subroutine RMX2 runs every frame at $7E2695 waits in a loop 3 different times for bit 6 of $7F5E, which is the CX4 running flag, to be 0. It waits for the CX4 to stop running, disables NMI and IRQs, starts the program preload, waits again, starts execution by writing to $7F4F, then waits again. Coprocessor fast mode appears to synchronize the CX4 once per scanline rather than once per CPU step. I can see why this would make it appear slower if the CX4 is flagged as running for the rest of an entire scanline before the CPU can continue. But it also makes me think there's a missing piece with regards to how and when the CX4 is no longer flagged as running. I wouldn't expect how often you sync to affect the timing that much, since it runs independently on real hardware anyway.

Screwtapello:
I wonder how accurate the timing of bsnes' CX4 emulation is. It looks like most instructions are assumed to be one cycle, except for a few that add two more, which seems... optimistic?

Making every CX4 operation take at least a scanline might be the kind of bug that cancels out inaccurate timing. I'm not sure if Near ever found a datasheet with timing information for the CX4, it might require additional research.

carmiker:
One more thing of note: changing the coprocessor settings also determines how the "attract mode" with the intro level boss works. I don't have my MMX2 cartridge handy right now to test on hardware, but you will notice that X will jump around erratically and die unless "Fast mode" is on. He also dies with HLE mode on.

AkiteruSDA:
> I wonder how accurate the timing of bsnes' CX4 emulation is. It looks like most instructions are assumed to be one cycle, except for a few that add two more, which seems... optimistic?
> 
> Making every CX4 operation take at least a scanline might be the kind of bug that cancels out inaccurate timing. I'm not sure if Near ever found a datasheet with timing information for the CX4, it might require additional research.

According to [ikari's tests](https://forums.nesdev.org/viewtopic.php?t=14647), those timings seem to be the case, though I can't test that for myself to confirm. It certainly would be nice to have a datasheet though.

> One more thing of note: changing the coprocessor settings also determines how the "attract mode" with the intro level boss works. I don't have my MMX2 cartridge handy right now to test on hardware, but you will notice that X will jump around erratically and die unless "Fast mode" is on. He also dies with HLE mode on.

Yeah, that seems to be a common issue if CX4 timing is off. I know Mesen and previously MiSTer had the same thing. To note, the attract mode boss plays as expected with both fast mode on/off with these changes.

I pushed a couple of changes that I think were inaccurate before. With these changes, the wireframe is much closer to console with fast mode off ([comparison](https://www.youtube.com/watch?v=ynfjThgBGwg)). It's still two frames slower, so maybe there's some very minor thing that can be tweaked somewhere else, but it's much closer than it was before with fast mode off.

I fixed what seems like an off-by-one error on each bus read/write instruction. According to ikari's tests, bus reads/writes should be 1+waitstates, but it looks like there was an extra cycle happening in bsnes:
```
assume WS = 3
LD -> set pending, pending = 1 + WS = 4
step(1) -> decrement pending, pending = 3
INC -> sees pending == 3
step(1) -> decrement pending, pending = 2
WAIT -> step(2), set pending = 0
step(1) -> pending == 0, data read
next instruction

total cycles until next instruction are 5, expected 4

```

I also added another `step(wait(address));` when reading the second byte per loop during cache loading. It seems likely that both of those reads require waiting, as there are no other instances of a 16-bit read happening while only having to wait once.

Screwtapello:
Looks good to me, thanks!

joepogo:
Is it possible to add this to the retroarch core bsnes v 115 core? 😊

Morilli:
Since these changes, [this cx4 test rom](https://archive.org/download/tukuyomi-snes-archive/bsnes-archive-tukuyomi.zip/nikel.me%2F~jiefk%2Fsnesemu%2Fmisc%2Froms%2Fcx4test.zip) no longer boots correctly, so something must be wrong or missing here. As far as I'm aware that test rom works correctly on hardware.

joepogo:
I just tested this test rom on the current retroarch version of bsnes v115 with the accuracy core and all tests passed. With the less accurate settings only a couple tests pass. 

I also tried the bsnesjg retroarch core and it's just a black screen. I'm assuming you're getting the same issue?

Morilli:
> I also tried the bsnesjg retroarch core and it's just a black screen. I'm assuming you're getting the same issue?

Yes, it's now just a black screen and doesn't display any test results. From my debugging it seems that the cpu is now trying to read from the dsp rom while the caching is in progress, which gets blocked by this line of code (during caching, `busy()` is true because `io.cache.enable` is set): https://github.com/bsnes-emu/bsnes/blob/7d5aa1e656b9171524d01b1b22917197d8121cb4/bsnes/sfc/coprocessor/hitachidsp/memory.cpp#L43

This could be fixed in numerous ways, for example by adding `|| io.cache.preload` to allow this specific case, but its unclear what the correct behavior is. I don't have hardware to experiment with so I can't verify that the test rom actually does work consistently on hardware or modify the test rom and do tests with that.

AkiteruSDA:
That appears to be an issue with the old test ROM where it writes to 7f48 to test the value without taking into account that it triggers a cache load. The test ROM doesn't have any CX4 code to load anyway, so it's an illegal operation. This patched ROM runs on the latest nightly: https://github.com/AkiteruSDA/cx4test/raw/refs/heads/main/cx4test.sfc

image

PR #392, opened 2026-08-02
Make our own "is_integral" and "is_signed" traits.

Screwtapello:
It seems that a lot of compilers implement a 128-bit integer type, but don't hook it up to all the standard library definitions. Previously, nall tried to amend the standard library definitions to match, but although that worked, it wasn't supposed to, and clang++ 20 apparently complains.

Although nothing in bsnes actually uses 128-bit integers for anything, it seems a shame to lose the functionality entirely, so we make our own traits, and adapt the serialization code to use it.

This patch was adapted from ares' commit 6a253b1.


Screwtapello:
It looks like the FreeBSD build failure is because of missing `<cairo.h>`, not related to this problem.

